PR #7, opened 2016-10-21
Merge rel/1.0.0-prevew back to master

xiaomi7732:
Merge rel/1.0.0-preview back to master branch. Preparing for the next release.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/project.json
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/project.json
@@ -20,6 +20,6 @@
   },
 
   "authors": [ "Microsoft" ],
-  "copyright": "© Microsoft Corporation. All rights reserved.",
+  "copyright": "Â© Microsoft Corporation. All rights reserved.",
   "description": "Provides an output implementation that sends diagnostics data to Microsoft Operations Management Suite."
 }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/project.json
```diff
@@ -20,6 +20,5 @@
   },
 
   "authors": [ "Microsoft" ],
-  "summary": "Operations Management Suite output for Microsoft EventFlow",
```
karolz-ms: Looks like you just removed the summary here. Was this intentional?

src/Microsoft.Diagnostics.EventFlow.Outputs.TableStorage/project.json
```diff
@@ -16,6 +16,5 @@
     "delaySign": true
   },
   "authors": [ "Microsoft" ],
-  "summary": "Azure Table Storage output for Microsoft Diagnostics EventFlow",
```
karolz-ms: Also here--is the removal of the summary intentional?

xiaomi7732: Yes, they are removed internally because they are causing build warnings when they are set out of the packOptions section.

PR #9, opened 2016-10-28
Elevate PerformanceCounter input to preview quality

karolz-ms:
1. Make PerformanceCounter input preview-ready
2. A couple of minor bug fixes for ElasticSearch output and Service Fabric platform support
3. Add capability to build Debug version of Nuget packages to our Nuget build script

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,59 @@ This input listens to EventSource traces. EventSource classes can be created in 
 | level | Critial, Error, Warning, Informational, Verbose, LogAlways | No | Specifies the collection trace level. Traces with equal or higher severity than specified are collected. For example, if Warning is specified, then Critial, Error, and Warning traces are collected. Default is All. |
 | keywords | An integer | No | A bitmask that specifies what events to collect. Only events with keyword matching the bitmask are collected, except if it's 0, which means everything is collected. Default is 0. |
 
+#### PerformanceCounter
+*Nuget Package*: **Microsoft.Diagnostics.EventFlow.Inputs.PerformanceCounter**
+
+This input enables gathering data from Windows performance counters. Only *process-specific* counters are supported, that is, the counter must have an instance that is associated with the current process. For machine-wide counters use an external agent such as [Azure Diagnostics Agent](https://azure.microsoft.com/en-us/documentation/articles/azure-diagnostics/) or create a custom input.
+
+*Finding the counter instance that corresponds to the current process*
+
+In general there is no canonical way to find a performance counter instance that corresponds to current process. Two methods are commonly used in practice:
+
+- A special performance counter that provides instance name to process ID mapping.
+    This solution involves a set of counters that use the same instance name for a given process. Among them there is a special counter with a value that is the process ID of the corresponding process. Searching for the instance of the special counter with a value equal to the current process ID allows to discover what the instance name is used for the current process. Examples of this approach include the Windows Process category (special counter "ID Process") and all .NET counters (special counter "Process ID" in the ".NET CLR Memory" category).
+    
+- Process ID can be encoded directly into the instance name.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -141,6 +141,66 @@ This input listens to EventSource traces. EventSource classes can be created in
 | level | Critial, Error, Warning, Informational, Verbose, LogAlways | No | Specifies the collection trace level. Traces with equal or higher severity than specified are collected. For example, if Warning is specified, then Critial, Error, and Warning traces are collected. Default is All. |
 | keywords | An integer | No | A bitmask that specifies what events to collect. Only events with keyword matching the bitmask are collected, except if it's 0, which means everything is collected. Default is 0. |
 
+#### PerformanceCounter
```
brahmnes: Thank you for adding this to readme.md!

src/Microsoft.Diagnostics.EventFlow.Inputs.PerformanceCounter/Configuration/PerformanceCounterConfiguration.cs
```diff
@@ -3,42 +3,87 @@
...
+using System;
+
 namespace Microsoft.Diagnostics.EventFlow.Configuration
 {
     // !!ACTION!!
     // If you make any changes here, please update the README.md file to reflect the new configuration
     public class PerformanceCounterConfiguration
     {
-        public string CounterCategory { get; set; }
+        private static readonly string DotNetProcessIdCounterCategory = ".NET CLR Memory";
```
brahmnes: Should add a comment explaining these counters are special that you need to special-map them.

src/Microsoft.Diagnostics.EventFlow.Inputs.PerformanceCounter/TrackedPerformanceCounter.cs
```diff
@@ -0,0 +1,77 @@
...
+        private string lastInstanceName;
+
+        public PerformanceCounterConfiguration Configuration { get; private set; }
+
+        public TrackedPerformanceCounter(PerformanceCounterConfiguration configuration)
+        {
+            Requires.NotNull(configuration, nameof(configuration));
+
+            this.Configuration = configuration;
+            this.lastAccessedOn = DateTimeOffset.UtcNow.Subtract(TimeSpan.FromDays(1));
```
karolz-ms: Comment added. The lastAccessedOn time is set way in the past so that the data is sampled at startup.

src/Microsoft.Diagnostics.EventFlow.Inputs.PerformanceCounter/TrackedPerformanceCounter.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var now = DateTimeOffset.UtcNow;
+            if (now - this.lastAccessedOn < TimeSpan.FromMilliseconds(this.Configuration.SamplingIntervalMsec))
+            {
+                return false;
+            }
+
+            string instanceName = processInstanceNameCache.GetCounterInstanceNameForCurrentProcess(this.Configuration);
+            this.lastAccessedOn = now;
+            if (this.counter == null || (!string.IsNullOrEmpty(instanceName) && instanceName != this.lastInstanceName))
```
karolz-ms: The PerformanceCounter instance naming scheme that is used by most counters is exename, exename#1, exename#2 and so on (assuming the interesting case, that is, multiple processes that are based on the same executable). But these instance names are not stable, meaning that as the processes start and stop, their instance names might change, because the system is trying to avoid "gaps" between instance names. Say you have two processes, first uses exename for counter instances, second uses exename#1. Then the first process ends. Now the second process will use exename for its instances.

It is kind of messy

PR #10, opened 2016-11-01
Add OMS (Operations Management Suite) output

karolz-ms:
The OMS guys were asking me to check out their new JSON-based ingestion endpoint, so I just went ahead and polished the OMS output prototype a bit. It works! :-)


karolz-ms:
@brahmnes Please approve the review if you are OK with the change. I know the documentation is still missing, I am working on it, should not take long

PR #12, opened 2016-11-07
Implement a contract for metric and request data retrieval

karolz-ms:
This PR started as an attempt to improve the handling of metric and request metadata in Elasticsearch output, but I quickly realized it would result in duplication of code between Elasticsearch and ApplicationInsights outputs. So I ended up implementing a contract for metric and request data retrieval and added it to the Core assembly.


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/RequestData.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+
+            string requestName = null;
+            if (!eventData.GetValueFromPayload<string>(requestNameProperty, (v) => requestName = v))
+            {
+                return DataRetrievalResult.DataMissingOrInvalid(requestNameProperty);
+            }
+
+            string isSuccessProperty = requestMetadata[IsSuccessPropertyMoniker];
+            if (string.IsNullOrWhiteSpace(isSuccessProperty))
```
yantang-msft: How about allow isSuccess as optional property?

yantang-msft: Actually I think even duration can be optional.
Anyways we didn't define response code, http method... which I think is at least as important as duration. So I don't think a request metadata without duration is invalid.

karolz-ms: @yantang-msft this is a good suggestion, thanks, Will make isSuccess and duration optional.

A note on HTTP method/response code: these are only valid for HTTP-based requests, so cannot be mandatory. I will add a response code property as optional. Will keep it as a string so that it is as general as possible, covering non-HTTP scenarios too.

PR #13, opened 2016-11-10
Improve Serilog input

karolz-ms:
1. Added documentation for Serilog input
2. Added some unit tests
3. Improved value extraction logic to get better type fidelity between what gets logged vs. what EventFlow pipeline is seeing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,70 @@ EventFlow PerformanceCounter input supports the first method of deterimining cou
 Some performance counters require the user to be a member of the Performance Monitor Users system group. 
 This can manifest itself by health reporter reporting "category does not exist" errors from PerformanceCounter output, 
 despite the fact that the category and counter are properly configured and clearly visible in Windows Performance Monitor. 
-If you need to consume such counters, make sure the account your process runs under belogs to Performance Monitor Users group.  
+If you need to consume such counters, make sure the account your process runs under belogs to Performance Monitor Users group.
+
+#### Serilog
+
+*Nuget package:* **Microsoft.Diagnostics.EventFlow.Inputs.Serilog**
+
+This input enables capturing diagnostic data created through [Serilog library](https://serilog.net/). It is designed to work with [Observable Serilog sink](https://github.com/serilog/serilog-sinks-observable). 
+
+*Configuration example*
+The Serilog input has no configuration, other than the "type" property that specifies the type of the input (must be "Serilog"):
+```json
+"inputs": [
+    {
+      "type": "Serilog",
+    }
+    // (other inputs, filters and outputs)
+```
+
+*Example: instantiating a Serilog logger that uses EventFlow Serilog input*
+
+eventFlowConfig.json:
+```json
+{
+  "inputs": [
+    {
+      "type": "Serilog",
+    }
+  ],
+  "outputs": [
+    {
+      "type": "StdOutput"
+    }
+  ]
+}
+```
+
+Program.cs:
+```csharp
+using System;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -199,7 +199,70 @@ EventFlow PerformanceCounter input supports the first method of deterimining cou
...
+#### Serilog
+
+*Nuget package:* **Microsoft.Diagnostics.EventFlow.Inputs.Serilog**
+
+This input enables capturing diagnostic data created through [Serilog library](https://serilog.net/). It is designed to work with [Observable Serilog sink](https://github.com/serilog/serilog-sinks-observable). 
+
+*Configuration example*
+The Serilog input has no configuration, other than the "type" property that specifies the type of the input (must be "Serilog"):
+```json
+"inputs": [
```
yantang-msft: I think the "inputs" [] part is not needed, which is consistent with other inputs/outputs. User should know to put the config in the inputs section. And the below config example is not needed as well.

README.md
```diff
@@ -199,7 +199,70 @@ EventFlow PerformanceCounter input supports the first method of deterimining cou
...
+namespace SerilogEventFlow
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            using (var pipeline = DiagnosticPipelineFactory.CreatePipeline(".\\eventFlowConfig.json"))
+            {
+                IObserver<LogEvent> serilogInput = pipeline.Inputs.OfType<SerilogInput>().First();
+                Log.Logger = new LoggerConfiguration().WriteTo.Observers(events => events.Subscribe(serilogInput)).CreateLogger();
```
yantang-msft: Don't know the details about Serilog. But should the Log.Logger be created by the Serilog input?
Ideally, user should just need one line of initialization and call Log.Info() however they want, just like the Trace input.

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/SerilogInput.cs
```diff
@@ -106,5 +106,42 @@ private EventData ToEventData(LogEvent logEvent)
...
+                    .Where(kvPair => kvPair.Key.Value is string && kvPair.Value is ScalarValue)
+                    .ToDictionary(kvPair => (string) kvPair.Key.Value, kvPair => ((ScalarValue) kvPair.Value).Value);
+                if (dictionaryResult.Count == dictionaryValue.Elements.Count)
+                {
+                    return dictionaryResult;
+                }
+            }
+
+            // Fall back to string rendering of the value
+            return logEventValue.ToString();
```
yantang-msft: nit: Just curious, is ToString() overridden for StructureValue?

karolz-ms: Yes, ToString() calls Render() which is overriden by StructureValue

README.md
karolz-ms: I do not think we should be in the business of creating Serilog loggers. They have a lot of different options and a specific, fluent-style API to handle them--we do not want to duplicate that. Also and perhaps more importantly, it is expected to create many context-specific "child" loggers during the execution of a service. So it is not like EventFlow pipeline that typically gets created at startup and exists till the service is shut down

README.md
karolz-ms: OK, I have removed the pieces you pointed out

README.md
yantang-msft: OK, if it's necessary to create child loggers on the fly, then Serilog can be a special case given its complexity.

PR #14, opened 2016-11-10
Implement event context

lmolkova:
This change implements event context for Trace and EventSource input - any state passed together with event to the output.
Context is captured in the input.
It's stored in the `EventData` payload.
Output implementations take care about it's format

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/Program.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/Program.cs
@@ -1,0 +1,75 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Diagnostics.Correlation.Common;
+using Newtonsoft.Json;
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp
+{
+    public class CorrelationTelemetryInitializer : ITelemetryInitializer
+    {
+        public void Initialize(ITelemetry telemetry)
+        {
+            //add request id to every event
+            var ctx = ContextResolver.GetRequestContext<MyContext>();
+            if (ctx != null)
+            {
+                telemetry.Context.Operation.Id = ctx.CorrelationId;
+            }
+        }
+    }
+
+    class MyContext
+    {
+        public string CorrelationId;
+        public string OtherId;
+
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder();
+            using (JsonWriter writer = new JsonTextWriter(new StringWriter(sb)))
+            {
+                writer.WriteStartObject();
+                writer.WritePropertyName(nameof(CorrelationId));
+                writer.WriteValue(CorrelationId);
+                writer.WritePropertyName(nameof(OtherId));
+                writer.WriteValue(OtherId);
+                writer.WriteEnd();
+                writer.WriteEndObject();
+            }
+            return sb.ToString();
+        }
+    }
+
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            TelemetryConfiguration.Active.TelemetryInitializers.Add(new CorrelationTelemetryInitializer());
+            TelemetryConfiguration.Active.TelemetryChannel.DeveloperMode = true;
+
+            using (DiagnosticPipeline pipeline = DiagnosticPipelineFactory.CreatePipeline("config.json"))
+            {
+                for (int i = 0; i < 10; i++)
+                {
+                    ContextResolver.SetRequestContext(new MyContext
+                    {
+                        CorrelationId = Guid.NewGuid().ToString(),
+                        OtherId = i.ToString()
+                    });
+                    Trace.TraceWarning($"{DateTime.UtcNow:o} this is log message");
+                }
+                Task.Delay(10000).Wait();
+            }
+        }
+    }
+}
--- a/src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/config.json
+++ b/src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/config.json
@@ -1,0 +1,28 @@
+﻿{
+  "healthReporter": {
+    "type": "CsvHealthReporter",
+    "logFileFolder": ".",
+    "logFilePrefix": "HealthReport",
+    "minReportLevel": "Warning",
+    "throttlingPeriodMsec": "1000"
+  },
+  "inputs": [
+    {
+      "type": "Trace",
+      "traceLevel": "Warning"
+    }
+  ],
+  "filters": [
+  ],
+  "outputs": [
+    {
+      "type": "ApplicationInsights",
+      "instrumentationKey": "**REMOVED**"
+    },
+    {
+      "type": "StdOutput"
+    }
+  ],
+
+  "schema-version": "2016-08-11"
+}
--- a/src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/Microsoft.Diagnostics.EventFlow.FilterParserGenerator.csproj
+++ b/src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/Microsoft.Diagnostics.EventFlow.FilterParserGenerator.csproj
@@ -32,6 +32,10 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Diagnostics.Correlation.Common, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Microsoft.Diagnostics.Correlation.Common.1.0.0-preview1\lib\net45\Microsoft.Diagnostics.Correlation.Common.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="Pegasus.Common">
       <HintPath>..\..\packages\Pegasus.3.1.2\lib\net40\Pegasus.Common.dll</HintPath>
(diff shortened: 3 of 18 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventDataExtensions.cs
```diff
@@ -105,9 +106,15 @@ private static ShoeBoxEventData ConvertToShoeboxMetric(EventData eventData, IRea
                     metricName = metadata["metricName"],
                     time = eventData.Timestamp,
                     timeGrain = null,
-                    dimensions = null,
+                    dimensions = new Dictionary<string, object>()
```
brahmnes: Is this fixing a bug that's unrelated to correlation?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventData.cs
```diff
@@ -28,6 +29,8 @@ public class EventData: IDeepCloneable<EventData>
 
         public long Keywords { get; set; }
 
+        public object EventContext { get; set; } = ContextResolver.GetRequestContext<object>();
```
brahmnes: I wonder if we should put this behind a flag. That is, by default, correlation functionality is turned off. If user turns it on in the config file, or through some switch on DiagnosticsPipeline, then we switch it on. The reason is that correlation usage has some, eventhough it's not significant, performance overhead.

lmolkova: If no context is set, then there is no impact on performance according to my measurements.
So, not setting context is silent way to opt out.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventData.cs
```diff
@@ -184,6 +191,8 @@ public EventData DeepClone()
             other.Level = this.Level;
             other.ProviderName = this.ProviderName;
             other.Timestamp = this.Timestamp;
+            other.EventContext = this.EventContext;
```
lmolkova: Agree

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventDataExtensions.cs
lmolkova: dimensions were not filled before, 
`metricRecord.dimensions.Add(nameof(eventData.EventContext), eventData.EventContext);` is called later

src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    class MyContext
+    {
+        public string CorrelationId;
+        public string OtherId;
+
+        public override string ToString()
+        {
+            return JsonConvert.SerializeObject(this);
```
karolz-ms: If you have just two string properties to serialize, serializing using JsonTextWriter (WriteStartObject, WritePropertyName, WriteValue, WriteEndObject) is much more efficient than JsonConvert.SerializeObject

src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/config.json
```diff
@@ -0,0 +1,39 @@
...
+  ],
+  "outputs": [
+    {
+      "type": "ApplicationInsights",
+      "instrumentationKey": "***REMOVED***"
+    },
+    {
+      "type": "StdOutput"
+    },
+    {
```
karolz-ms: What is the reason for including 4 different outputs in the configuration? Would AI and StdOutput be sufficient?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventData.cs
karolz-ms: The problem with event context, as it is defined today, is that it is an "object". So we do not know its cloning semantics.

If the context is immutable, the current implementation is correct. If it is not, and it might be changing, then we should clone it, but because it is an object, we do not how.

One possibility is to check if the context implements IDeepCloneable. If it does, clone it. If not, do what current implementation does.

src/Microsoft.Diagnostics.EventFlow.Core/project.json
```diff
@@ -1,6 +1,7 @@
 {
   "version": "1.0.0-*",
   "dependencies": {
+    "Microsoft.Diagnostics.Correlation.Common": "1.0.0-preview1",
```
karolz-ms: I was planning to move EventFlow out of preview some time towards the end of November and start using it in Service Fabric templates. For that, we should not rely on any preview libraries. How close is the Correlation lib from crossing the boundary from preview to "production"?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventData.cs
karolz-ms: Depending on the EventFlow input implementation, there might be no context when the EventData is being created, or the context might be different from the context that was valid when original event was raised. 

A safer design would be to have the input fill the context. The implication is that EventData constructor probably does not need to set the context

karolz-ms: This design also has the advantage that EventFlow Core library does not need to depend on the Correlation library anymore

lmolkova: The problem is that not all inputs support any kind of context. 
1. Trace: no context
2. EventSource: has payload, but it's mapped to EventData.Payload
3. PerformanceCounter: context does not make sense
4. Serilog proprieties are implemented as payload
5. ILogger supports state and consistent approach would be to pass it in the payload

It makes me think that 
1. for most inputs (which support some kind of context), payload is existing way to pass it
2. `TraceInput` may get the context with `ContextResolver.GetContext<object>()` and put it in the payload, But I really doubt that we should do it for the inputs which do not have any context by it's nature.

src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/config.json
lmolkova: It was needed to check how context is serialized to other outputs. However as a sample it does not need to write to all outputs and I left StdOut and AI only

src/Microsoft.Diagnostics.EventFlow.Core/project.json
lmolkova: as discussed, we will be able to provide release version with limited API

src/Microsoft.Diagnostics.EventFlow.Consumers.EventContextConsoleApp/Program.cs
lmolkova: changed to JsonTextWriter

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventData.cs
lmolkova: Context is now only filled in Trace and EventSource input. AppInsights output also need to restore context, so it depends on Correlation lib

src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/Microsoft.Diagnostics.EventFlow.FilterParserGenerator.csproj
```diff
@@ -32,6 +32,10 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Diagnostics.Correlation.Common, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
kzadora: Is it really necessary for the FilterParserGenerator to take a dependency on Correlation lib?

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs
```diff
@@ -29,6 +30,11 @@ public static EventData ToEventData(this EventWrittenEventArgs eventSourceEvent)
             eventPayload["EventId"] = eventSourceEvent.EventId;
             eventPayload["EventName"] = eventSourceEvent.EventName;
             eventPayload["ActivityID"] = ActivityPathString(eventSourceEvent.ActivityId);
+            var eventContext = ContextResolver.GetRequestContext<object>();
+            if (eventContext != null)
+            {
+                eventPayload["EventContext"] = eventContext;
```
kzadora: 1. Is "RequestContext" a better name?
2. What if I want to set the request context at the time when the event is created (not at the time when it is processed by the input)?

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
```diff
@@ -200,6 +201,12 @@ private void SubmitEventData(string message, TraceEventType level, int? id = nul
                     eventPayload["RelatedActivityID"] = relatedActivityId;
                 }
 
+                var eventContext = ContextResolver.GetRequestContext<object>();
+                if (eventContext != null)
+                {
+                    eventPayload["EventContext"] = eventContext;
```
kzadora: Same questions as for the EventSource input apply here too

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs
lmolkova: 1. It's not necessary request context. It's any context related to this Trace, so "Request" might be misleading
2. Since Trace API does not allow to pass it, I believe we cannot achieve it with trace input.  Could you elaborate?

lmolkova: 1. Oh, then this API is misleading. Let me fix this API as a part of Correlation release process, and stick with `EventContext` here.

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
karolz-ms: Given what Vance said in the other conversation (about OnEventWritten being called synchronously) I think what you have is fine. That is, it is not possible that the context will change between the time EventSource event method is called and when OnEventWritten/EventSource input is called

PR #15, opened 2016-11-11
Add documentation for standard metadata types

karolz-ms:
Also change formatting of the field names in tables to make it a bit more visually appealing.


yantang-msft:
Thanks for this change!

karolz-ms:
Thanks for the review, Yanming!

PR #18, opened 2016-11-17
ILogger input

lmolkova:
This change implements ILogger input

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,48 @@ namespace SerilogEventFlow
 }
 ```
 
+#### ILogger
+
+*Nuget package:* **Microsoft.Diagnostics.EventFlow.Inputs.ILogger**
+
+This input enables capturing diagnostic data created through Microsoft.Extensions.Logging library and ILogger interface. 
+
+*Configuration example*
+The ILogger input has no configuration, other than the "type" property that specifies the type of the input (must be "ILogger"):
+```json
+{
+  "type": "ILogger"
+}
+```
+
+*Example: instantiating a ILogger that uses EventFlow ILogger input*
+
+```csharp
+using Microsoft.Diagnostics.EventFlow;
+using Microsoft.Extensions.Logging;
+
+namespace LoggerEventFlow
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            using (var pipeline = DiagnosticPipelineFactory.CreatePipeline(".\\eventFlowConfig.json"))
+            {
+                var factory = new LoggerFactory()
+                    .AddEventFlow(pipeline);
+
+                var logger = new Logger<Program>(factory);
+                using (logger.BeginScope(myState))
+                {
+                    logger.LogInformation(""Hello from {friend} for {family}!", "LoggerInput", "EventFlow");
+                }
+            }
+        }
+    }
+}
+```
+
 ### Outputs
 Outputs define where data will be published from the engine. It's an error if there are no outputs defined. Each output type has its own set of parameters.
 
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ This input enables capturing diagnostic data created through Microsoft.Extension
 The ILogger input has no configuration, other than the "type" property that specifies the type of the input (must be "ILogger"):
 ```json
 {
-  "type": "ILogger"
+  "type": "Microsoft.Extensions.Logging"
 }
 ```
 
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
@@ -1,0 +1,111 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Internal;
+
+namespace Microsoft.Diagnostics.EventFlow.Inputs.ILogger.Logger
+{
+    internal class EventFlowLogger : Extensions.Logging.ILogger
+    {
+        private readonly string categoryName;
+        private readonly LoggerInput loggerInput;
+        private Scope<object> scope;
+
+        public EventFlowLogger(string categoryName, LoggerInput loggerInput)
+        {
+            Debug.Assert(categoryName != null);
+            Debug.Assert(loggerInput != null);
+
+            this.categoryName = categoryName;
+            this.loggerInput = loggerInput;
+        }
+
+        public void Log<TState>(Extensions.Logging.LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+                return;
+            var message = formatter(state, exception);
+
+            Dictionary<string, object> properties = new Dictionary<string, object>();
+            if (state is FormattedLogValues)
+            {
+                var formattedState = state as FormattedLogValues;
+                //last KV is the whole message, we will pass it separately
+                for (int i = 0; i < formattedState.Count - 1; i++)
+                    properties.Add(formattedState[i].Key, formattedState[i].Value);
+            }
+            if (scope?.State != null)
+            {
+                var formattedState = scope.State as FormattedLogValues;
+                if (formattedState != null)
+                {
+                    for (int i = 0; i < formattedState.Count - 1; i++)
+                    {
+                        if (!properties.ContainsKey(formattedState[i].Key)) //message and scope may have the same property
+                            properties.Add(formattedState[i].Key, formattedState[i].Value);
+                    }
+                    //last KV is the whole 'scope' message, we will add it formatted
+                    properties.Add("Scope", formattedState.ToString());
+                }
+                else
+                {
+                    properties.Add("Scope", scope.State);
+                }
+            }
+
+            loggerInput.SubmitEventData(message, ToLogLevel(logLevel), eventId, exception, categoryName, properties);
+        }
+
+        public bool IsEnabled(Extensions.Logging.LogLevel logLevel)
+        {
+            return true;
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            if (state == null)
+                return null;
+            scope = new Scope<object>(state);
+            return scope;
+        }
+
+        private class Scope<TState> : IDisposable
+        {
+            public TState State { get; private set; }
+
+            public Scope(TState state)
+            {
+                this.State = state;
+            }
+
+            public void Dispose()
+            {
+                var disposable = State as IDisposable;
+                disposable?.Dispose();
+                State = default(TState);
+            }
+        }
+
+        private LogLevel ToLogLevel(Extensions.Logging.LogLevel loggerLevel)
+        {
+            switch (loggerLevel)
+            {
+                case Extensions.Logging.LogLevel.Critical:
+                    return LogLevel.Critical;
+                case Extensions.Logging.LogLevel.Error:
+                    return LogLevel.Error;
+                case Extensions.Logging.LogLevel.Warning:
+                    return LogLevel.Warning;
+                case Extensions.Logging.LogLevel.Information:
+                    return LogLevel.Informational;
+            }
+            return LogLevel.Verbose;
+        }
+    }
+}
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/LoggerInput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/LoggerInput.cs
@@ -1,0 +1,73 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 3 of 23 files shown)


README.md
```diff
@@ -245,6 +245,48 @@ namespace SerilogEventFlow
 }
 ```
 
+#### ILogger
+
+*Nuget package:* **Microsoft.Diagnostics.EventFlow.Inputs.ILogger**
```
karolz-ms: I is likely that a lot of different libraries will use the ILogger concept and define their own ILogger implementation. For example Serilog has one: https://github.com/serilog/serilog/blob/dev/src/Serilog/ILogger.cs

So I suggest that we name the package and the input so that it conveys it is about Microsoft.Extensions.Logging package and its implementation of ILogger. For example

Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging

README.md
```diff
@@ -245,6 +245,48 @@ namespace SerilogEventFlow
...
+
+This input enables capturing diagnostic data created through Microsoft.Extensions.Logging library and ILogger interface. 
+
+*Configuration example*
+The ILogger input has no configuration, other than the "type" property that specifies the type of the input (must be "ILogger"):
+```json
+{
+  "type": "ILogger"
```
karolz-ms: How about using "Microsoft.Extensions.Logging" as the input type?

README.md
```diff
@@ -245,6 +245,48 @@ namespace SerilogEventFlow
...
+namespace LoggerEventFlow
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            using (var pipeline = DiagnosticPipelineFactory.CreatePipeline(".\\eventFlowConfig.json"))
+            {
+                var factory = new LoggerFactory()
+                    .AddEventFlow(pipeline);
```
karolz-ms: I like this. Very clean!

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Internal;
+
+namespace Microsoft.Diagnostics.EventFlow.Inputs.ILogger.Logger
```
karolz-ms: We use just Microsoft.Diagnostics.EventFlow.Inputs namespace for all input code

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    internal class EventFlowLogger : Extensions.Logging.ILogger
+    {
+        private readonly string categoryName;
+        private readonly LoggerInput loggerInput;
+        private Scope<object> scope;
+
+        public EventFlowLogger(string categoryName, LoggerInput loggerInput)
+        {
+            Debug.Assert(categoryName != null);
```
karolz-ms: Use Validation.NotNull() for parameter checking

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+            Debug.Assert(loggerInput != null);
+
+            this.categoryName = categoryName;
+            this.loggerInput = loggerInput;
+        }
+
+        public void Log<TState>(Extensions.Logging.LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
```
karolz-ms: Public method, so some parameter checking needs to take place

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+                var formattedState = scope.State as FormattedLogValues;
+                if (formattedState != null)
+                {
+                    for (int i = 0; i < formattedState.Count - 1; i++)
+                    {
+                        if (!properties.ContainsKey(formattedState[i].Key)) //message and scope may have the same property
+                            properties.Add(formattedState[i].Key, formattedState[i].Value);
+                    }
+                    //last KV is the whole 'scope' message, we will add it formatted
+                    properties.Add("Scope", formattedState.ToString());
```
karolz-ms: Can there be a "Scope" property on the scope? Should we check for duplicate property names too?

karolz-ms: or on the message?

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+                else
+                {
+                    properties.Add("Scope", scope.State);
```
karolz-ms: Same concern about duplicate property names here

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+
+            loggerInput.SubmitEventData(message, ToLogLevel(logLevel), eventId, exception, categoryName, properties);
+        }
+
+        public bool IsEnabled(Extensions.Logging.LogLevel logLevel)
+        {
+            return true;
```
karolz-ms: Hmm. Should we allow to drop events based on the level inside this Logger? Seems like a common requirement--during routine execution you probably want to capture only warnings or worse.

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
karolz-ms: (applies to other files too)

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/LoggerInput.cs
```diff
@@ -0,0 +1,73 @@
...
+                    eventPayload["EventName"] = eventId.Name;
+                }
+                if (exception != null)
+                {
+                    eventPayload["Exception"] = exception;
+                }
+                if (payload != null)
+                {
+                    foreach (var kv in payload)
+                        eventPayload.Add(kv);
```
karolz-ms: What if the payload already has properties named "Message", "EventId", "EventName" etc?

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/LoggerInput.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+
+                this.subject.OnNext(eventEntry);
+            }
+            catch (Exception ex)
+            {
+                this.healthReporter?.ReportProblem($"Failed to write message. Error: {ex}", TraceTag);
```
karolz-ms: 1. Health reporter will never be null (given constructor check)
2. Reporting a warning is probably more appropriate 
3. Use ex.ToString() for extra clarity

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/LoggerInputFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Diagnostics.EventFlow.Inputs
+{
+    public class LoggerInputFactory : IPipelineItemFactory<LoggerInput>
+    {
+        public LoggerInput CreateItem(IConfiguration configuration, IHealthReporter healthReporter)
+        {
+            Validation.Requires.NotNull(configuration, nameof(configuration));
```
karolz-ms: Right now configuration is not used so I am not sure if checking for null is entirely warranted. But if you decide to implement level-based event filtering, you will need the configuration

test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/LoggerInputTests.cs
```diff
@@ -0,0 +1,232 @@
+﻿using System;
```
karolz-ms: Nice job with the tests!!

README.md
cwe1ss: +1, this name is very confusing.

lmolkova: will rename to Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging

README.md
lmolkova: agree

README.md
lmolkova: thanks!

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
lmolkova: fixed

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
lmolkova: ok (EventFlowLogger is internal, so it's not public API)

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
lmolkova: We can check, but what to do with duplicate? 
I see two possible ways
1. Ignore duplicates and report error to health reporter.
2. Keep all properties from the message in separate dict under payload, but then LoggerInput would behave differently than EventSource or Serilog. 


EventSourceInput and Serilog seems to have the same problem. So may be all of them should be fixed in the same way?

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
lmolkova: Level based event filtering could be done without EventFlow when factory is being created:
```
factory
.WithFilter(new FilterLoggerSettings { { "ConsoleApp1.Program", LogLevel.Warning } })
.WithEventFlow();
```
If we additionally configure level-based filtering on EventFlow side it may conflict with MS.Extenstion.Logging approach.

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/LoggerInput.cs
lmolkova: will fix

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
karolz-ms: Understood, but even for internal types, I like to do parameter checking and throw if the checks fail. I like to think that public methods & properties form a contract with consumers and parameter checking is part of that contract. OK to just assert the correctness of parameters of private methods.

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
karolz-ms: Agreed that all inputs should behave the same way. Feel free to open an issue about this and we can deal with it as a separate change.

I think whatever the original event carries should take priority over what the input adds. So if there is a "Message" property on the event, we should not overwrite it. We could either skip input additions completely or make the added property name unique, e.g. by appending random number at the end of the name. These are, essentially, variations of the (1) option you have. I have slight preference for the latter. 

I do not think using a separate dictionary is a good idea. It will make the output code and, ultimately, consumption of the data much more complicated, for a marginal benefit.

Oh, and I would just report a warning, not an error, if a name collision happens, esp. if you decide to make the added property name unique and include the data, mitigating the conflict.

src/Microsoft.Diagnostics.EventFlow.Inputs.ILogger/Logger/EventFlowLogger.cs
karolz-ms: Sounds good, let's not have this extra configuration parameter

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/DictionaryExtenstions.cs
```diff
@@ -0,0 +1,57 @@
...
+            Validation.Requires.NotNull(dictionary, nameof(dictionary));
+            Validation.Requires.NotNull(key, nameof(key));
+
+            if (!dictionary.ContainsKey(key))
+            {
+                dictionary.Add(key, value);
+                return new AddResult();
+            }
+
+            Random random = new Random(DateTime.Now.Millisecond);
```
karolz-ms: 1. No need to use Random(seed) constructor; the parameter-less constructor does what your code does
2. How about creating a static instance of the Random class and caching it?

lmolkova: fixed

karolz-ms:
One more thing: in Warsaw.sln make sure the new MicrosoftLogging project is listed as a dependency of the signing project

Issue #16, opened 2016-11-15
Shutdown of DataPipeline may lead to data loss


karolz-ms:
Repro:

bool cont = true;

                Console.WriteLine("String Logger!");

                while (cont)
                {
                    Console.WriteLine();
                    Console.WriteLine("Enter any text to log. Entering 'X' will exit.");
                    string input = Console.ReadLine().Trim();

                    if (input.Equals("X", StringComparison.InvariantCultureIgnoreCase))
                    {
                        cont = false;
                    }
                    else
                    {
                        using (var pipeline = DiagnosticPipelineFactory.CreatePipeline("eventFlowConfig.json"))
                        {
                            
                            System.Diagnostics.Trace.TraceWarning($"TraceWarning: {input}");
                            Console.WriteLine($"{input} logged.");
                        }
                    }
                }

Expected -- if you run the program above with Trace input and StdOutput you should see the logs

Actual -- because the pipeline is destroyed immediately after the trace is submitted, no logs show up in the output

PR #26, opened 2016-12-14
Ensure diagnostic pipeline is drained before Dispose() returns

karolz-ms:
This solves issue https://github.com/Azure/diagnostics-eventflow/issues/16

Also included in this change:
1. An update to Nuget package creation script (using PowerShell will make it easier to maintain going forward)

NOTE: this will require changing the TFS build definitions

2. Make sure the TraceInput is unregistered from the tracing infrastructure when disposing


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -156,11 +184,14 @@ public void Dispose()
 
             DisposeOf(this.inputSubscriptions);
 
-            this.pipelineHead.Complete();
-            this.pipelineHead.Completion.Wait(TimeSpan.FromMilliseconds(this.pipelineConfiguration.PipelineCompletionTimeoutMsec));
+            TimeSpan pipelineDrainWaitTime = PollWaitForPipelineDrain();
+
+            pipelineHead.Complete();
+            TimeSpan completionWaitTime = TimeSpan.FromMilliseconds(Math.Max(100, this.pipelineConfiguration.PipelineCompletionTimeoutMsec - pipelineDrainWaitTime.TotalMilliseconds));
```
brahmnes: Should add a comment about why the 100 ms being the minimum waiting time.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -229,16 +290,19 @@ public EventData[] FilterAndDecorate(EventData[] eventsToFilter)
                     return null;
                 }
 
-                // CONSIDER: using a Queue<T> with external lock and be able to reuse the queue
-                // (ConcurrentQueue has no Clear() method while ordinary Queue does).
-                ConcurrentQueue<EventData> eventsToKeep = new ConcurrentQueue<EventData>();
+                ConcurrentStack<EventData> eventsToKeep;
```
brahmnes: Why switch from queue to stack? With the switch, the events are now returned in reverse order in the array. Does it matter?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -188,24 +219,52 @@ private void DisposeOf(IEnumerable<object> items)
...
+        }
+
         private class FilterAction
         {
             private IReadOnlyCollection<IFilter> filters;
             private ParallelOptions parallelOptions;
             private ExecutionDataflowBlockOptions executionDataflowBlockOptions;
             private IHealthReporter healthReporter;
+            private Action<int> eventsRemovedNotification;
+            private ConcurrentBag<ConcurrentStack<EventData>> stackPool;
```
yantang-msft: I think this deserves a comment

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -13,18 +13,20 @@
...
     {
-        private List<IDisposable> pipelineDisposables;
+        private List<IDisposable> pipelineLinkDisposables;
         private List<IDisposable> inputSubscriptions;
         private bool disposed;
         private bool disposeDependencies;
         private CancellationTokenSource cancellationTokenSource;
         private IDataflowBlock pipelineHead;
         private DiagnosticPipelineConfiguration pipelineConfiguration;
+        private volatile int eventsInProgress;
```
yantang-msft: I feel it's hard to maintain this variable across blocks. Ideally it's supported by DataFlow natively.
Is it possible to have such a variable for each block, and dispose from the head to the tail?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -97,14 +110,24 @@ public class DiagnosticPipeline : IDisposable
 
             if (sinks.Count > 1)
             {
+                // After broadcasting we will effectively have (sinks.Count - 1) more events in the pipeline.
+                var eventCounter = new TransformBlock<EventData[], EventData[]>(
+                    (batch) => { Interlocked.Add(ref this.eventsInProgress, (sinks.Count - 1) * batch.Length);  return batch; },
```
yantang-msft: Can you add more comment on this?
This is what I see, and it's even.
NumObservedAtInput - NumDroppedAtFilter - NumOutput = 0

How does this eventCounter work in the calculation of eventInProgress?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
karolz-ms: I will add a comment. 

It is not just about draining the pipeline from events, but also about disposing of all internal resources that the pipeline blocks may use. Even if the pipeline is not fully drained (e.g. due to stuck filter/output), we still want to call Complete and give it some time.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
karolz-ms: We do not guarantee the order of the events in the pipeline will be maintained, so from correctness perspective the change is fine. Actually, there is no guarantee that events will be submitted to the pipeline in any particular order. Nor do we guarantee that the batches of events are sent out in order by outputs.

Each event has a timestamp so we know what happened when.

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
```diff
@@ -161,15 +162,35 @@ public IDisposable Subscribe(IObserver<EventData> observer)
...
         {
+            if (!this.disposed)
+            {
+                lock (this.subject)
+                {
+                    if (!this.disposed)
+                    {
+                        try
+                        {
+                            this.disposed = true;
```
karolz-ms: Oops. Fixed

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
```diff
@@ -161,15 +162,35 @@ public IDisposable Subscribe(IObserver<EventData> observer)
...
+                        }
+                        finally
+                        {
+                            this.disposed = true;
+                        }
+                    }
+                }
+            }
+
             base.Dispose(disposing);
```
karolz-ms: According to .NET design guidelines one should be able to call Dispose() multiple times on the same object, with no ill effects. In this particular piece of code we want to ensure that we call Trace.Listeners.Remove(this) _exactly once_ as part of disposing, regardless how many times Dispose() is called

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
karolz-ms: Ideally, yes.

If we do not use completion propagation, complete from head to tail, and wait for completion on each block, that would give us what we want, i.e. drain the pipeline ensuring that each event is processed fully.

But that would be a bit tricky code to walk the pipeline and also account for how much time is left before we hit the timeout.

I think counting the events in the pipeline is simpler.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -40,16 +42,26 @@ public class DiagnosticPipeline : IDisposable
...
             Requires.Argument(sinks.Count > 0, nameof(sinks), "There must be at least one sink");
 
+            this.eventsInProgress = 0;
+
             this.pipelineConfiguration = pipelineConfiguration ?? new DiagnosticPipelineConfiguration();
 
+            // An estimatie how many batches of events to allow inside the pipeline.
+            // We want to be able to process full buffer of events, but also have enough batches in play in case of high concurrency.
+            int MaxNumberOfBatchesInProgress = Math.Max(
+                5 * this.pipelineConfiguration.MaxConcurrency,
```
karolz-ms: Yes, just a guess for reasonable size of the input buffer for the blocks that handler event batches.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -116,25 +139,30 @@ public class DiagnosticPipeline : IDisposable
...
                     sink.Output,
                     this.cancellationTokenSource.Token,
-                    this.pipelineConfiguration.MaxEventBatchSize,
+                    MaxNumberOfBatchesInProgress,
                     this.pipelineConfiguration.MaxConcurrency,
-                    healthReporter);
-                ActionBlock<EventData[]> outputBlock = outputAction.GetOutputBlock();
-                this.pipelineDisposables.Add(outputSource.LinkTo(outputBlock, propagateCompletion));
+                    healthReporter,
+                    (eventsSentCount) => Interlocked.Add(ref this.eventsInProgress, -eventsSentCount));
```
karolz-ms: Interlocked method do not use any locks.

I run the benchmark and there is no discernible effect on performance.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -249,7 +313,15 @@ public EventData[] FilterAndDecorate(EventData[] eventsToFilter)
...
                 });
 
-                return eventsToKeep.ToArray();
+                EventData[] eventsFiltered = eventsToKeep.ToArray();
+                eventsToKeep.Clear();
+                this.stackPool.Add(eventsToKeep);
+                int eventsRemoved = eventsToFilter.Length - eventsFiltered.Length;
+                if (eventsRemoved > 0)
+                {
+                    this.eventsRemovedNotification(eventsRemoved);
```
karolz-ms: That is ensured in the constructor

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
karolz-ms: The broadcaster clones the events, that is why we need to increase the counter. Once you realize that it is actually pretty simple. I have updated the comment to make it clearer.

PR #34, opened 2017-01-05
Make the Elasticsearch output work with Elasticsearch 5

karolz-ms:
1. Setting Refresh on BulkRequest is actually an anti-pattern, so I will probably apply that change to the 1.0 release as well. 
2. The plan is to release version 2.0 of the ES output Nuget based on this change. Version 1.0 Nuget will keep backward compat. with ES 2.x


karolz-ms:
Thanks!

PR #35, opened 2017-01-11
Convert `StructureValue` into `IDictionary<string,object>` when possible

nblumhardt:
Howdy! Interesting project. I just had a look at the Serilog input and noticed `StructureValue`, which is the core key/value pair representation in Serilog properties, wasn't yet handled. I've taken a cue from how `DictionaryValue` is handled, mapping the value onto raw `Dictionary<string, object>`.

I gather from the codebase that arbitrarily-deep structured data isn't allowed, hence the fallback to `ToString()` when an element of an array/dictionary is non-scalar. Rather than `ToString()` the whole value, the problematic element itself can be stringified, leaving the overall structure intact.

Keen to know your thoughts, to better understand where the code for the input is heading. I can see a few more possible improvements to the code and tests for the Serilog input, which I'd be happy to follow up on if I'm on the right track with this PR.


nblumhardt:
Great, thanks Karol.

PR #40, opened 2017-02-02
Streamline Serilog input configuration

nblumhardt:
This PR aligns Serilog input configuration more closely with typical Serilog syntax by adding `WriteTo.EventFlow()` and breaking the dependency on _Serilog.Sinks.Observable_:

```csharp
using System;
using System.Linq;
using Microsoft.Diagnostics.EventFlow;
using Serilog;

namespace SerilogEventFlow
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var pipeline = DiagnosticPipelineFactory.CreatePipeline(".\\eventFlowConfig.json"))
            {
                Log.Logger = new LoggerConfiguration()
                    .WriteTo.EventFlow(pipeline)
                    .CreateLogger();

                Log.Information("Hello from {friend} for {family}!", "SerilogInput", "EventFlow");
                
                Log.CloseAndFlush();
                Console.ReadKey();
            }
        }
    }
}
```

Though the `WriteTo.EventFlow()` shortcut accepts a `DiagnosticPipeline` and performs the search for the first `SerilogInput` itself, it's still possible to add a specific `SerilogInput` directly by passing it through `WriteTo.Sink()`.

The PR also reverses the priority of the built-in Message and Exception properties to take precedence over any attached properties with the same name. It's not unusual for Serilog events to carry serialized message payloads (e.g. received from some kind of service bus) as `Message`, along with the typical text message.

Later in processing, it will be less surprising to get the text message in `Message` (consistently with other events) and the payload object in `Message_1` or similar. (Happy to back this out of course, but it seems like a worthwhile change to make.)

Keen for any feedback/suggestions. Cheers!

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
@@ -1,9 +1,14 @@
-﻿using Microsoft.Diagnostics.EventFlow;
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using Microsoft.Diagnostics.EventFlow;
 using Microsoft.Diagnostics.EventFlow.Inputs;
 using Serilog.Configuration;
 using Serilog.Debugging;
-using System;
 using System.Linq;
+using Validation;
 
 namespace Serilog
 {
@@ -24,8 +29,8 @@ namespace Serilog
         /// instead.</remarks>
         public static LoggerConfiguration EventFlow(this LoggerSinkConfiguration loggerSinkConfiguration, DiagnosticPipeline diagnosticPipeline)
         {
-            if (loggerSinkConfiguration == null) throw new ArgumentNullException(nameof(loggerSinkConfiguration));
-            if (diagnosticPipeline == null) throw new ArgumentNullException(nameof(diagnosticPipeline));
+            Requires.NotNull(loggerSinkConfiguration, nameof(loggerSinkConfiguration));
+            Requires.NotNull(diagnosticPipeline, nameof(diagnosticPipeline));
 
             var input = diagnosticPipeline.Inputs.OfType<SerilogInput>().FirstOrDefault();
 
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
@@ -1,0 +1,41 @@
+﻿using Microsoft.Diagnostics.EventFlow;
+using Microsoft.Diagnostics.EventFlow.Inputs;
+using Serilog.Configuration;
+using Serilog.Debugging;
+using System;
+using System.Linq;
+
+namespace Serilog
+{
+    /// <summary>
+    /// Extends <see cref="LoggerSinkConfiguration"/>
+    /// </summary>
+    public static class LoggerSinkConfigurationEventFlowExtensions
+    {
+        /// <summary>
+        /// Publish Serilog events through a <see cref="DiagnosticPipeline"/>.
+        /// </summary>
+        /// <param name="loggerSinkConfiguration">The <c>WriteTo</c> object exposed by <see cref="LoggerConfiguration"/>.</param>
+        /// <param name="diagnosticPipeline">A configured <see cref="DiagnosticPipeline"/>. At least one Serilog input
+        /// must be configured for the pipeline.</param>
+        /// <returns>A <see cref="LoggerConfiguration"/> allowing configuration to continue.</returns>
+        /// <remarks>The method will select the first <see cref="SerilogInput"/> in the pipeline. If a specific input is desired instead,
+        /// this can be passed to <see cref="LoggerSinkConfiguration.Sink(Core.ILogEventSink, Events.LogEventLevel, Core.LoggingLevelSwitch)"/>
+        /// instead.</remarks>
+        public static LoggerConfiguration EventFlow(this LoggerSinkConfiguration loggerSinkConfiguration, DiagnosticPipeline diagnosticPipeline)
+        {
+            if (loggerSinkConfiguration == null) throw new ArgumentNullException(nameof(loggerSinkConfiguration));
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <param name="loggerSinkConfiguration">The <c>WriteTo</c> object exposed by <see cref="LoggerConfiguration"/>.</param>
+        /// <param name="diagnosticPipeline">A configured <see cref="DiagnosticPipeline"/>. At least one Serilog input
+        /// must be configured for the pipeline.</param>
+        /// <returns>A <see cref="LoggerConfiguration"/> allowing configuration to continue.</returns>
+        /// <remarks>The method will select the first <see cref="SerilogInput"/> in the pipeline. If a specific input is desired instead,
+        /// this can be passed to <see cref="LoggerSinkConfiguration.Sink(Core.ILogEventSink, Events.LogEventLevel, Core.LoggingLevelSwitch)"/>
+        /// instead.</remarks>
+        public static LoggerConfiguration EventFlow(this LoggerSinkConfiguration loggerSinkConfiguration, DiagnosticPipeline diagnosticPipeline)
+        {
+            if (loggerSinkConfiguration == null) throw new ArgumentNullException(nameof(loggerSinkConfiguration));
```
karolz-ms: (minor) We already take a dependency on the Validation library elsewhere in EventFlow code, so the null checks can be simplified to Requires.NotNull(inputParameter, nameof(inputParameter)

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            if (diagnosticPipeline == null) throw new ArgumentNullException(nameof(diagnosticPipeline));
+
+            var input = diagnosticPipeline.Inputs.OfType<SerilogInput>().FirstOrDefault();
+
+            if (input == null)
+            {
+                SelfLog.WriteLine("{0}: A Serilog input has not been added to the diagnostic pipeline; events will not be published to EventFlow", nameof(LoggerSinkConfigurationEventFlowExtensions));
+                return loggerSinkConfiguration.Sink(new LoggerConfiguration().CreateLogger()); // A "null" sink.
```
karolz-ms: Just curious: I assume returning a "do nothing" sink is a Serilog convention and preferable to, say, throwing an exception here?

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Microsoft.Diagnostics.EventFlow;
```
karolz-ms: If you do not mind, can you add a comment saying this file is licensed under MIT license?

karolz-ms:
Nicholas, thank you for your contribution, much appreciated. Overall looks very good and I am glad that you are helping us to make the Serilog input more "idiomatic".

The thinking behind preferring payload property names to built-in property names was that if the user of the library wants to have a "Message" payload property, the library should get out of the way. Please correct me if I am wrong, but you seem to be saying that a much more common situation is to have a conflict when the event is carrying payload from some other system and that the built-in property names should be preferred?

If that is the case, we might adopt this rule for the whole library. (I am not asking you to make the change as part of this PR, just thinking out loud).

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
nblumhardt: Done; just copied in the header from the other files 👍

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
nblumhardt: Missed that - thanks.

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
nblumhardt: I initially had an exception thrown here, but then considered that since the pipeline is configured using JSON, it's possible users might want to enable the Serilog input in some environment, and omit it in another. I was concerned that throwing an exception might introduce friction for this use case. Keen to fit with the rest of EventFlow so happy either way if you think an exception would be less surprising.

nblumhardt:
Thanks for the review.

The `Message` situation in Serilog usually looks like a variation of this:

```csharp
Log.Information("Received {Message}", message);
```

where `message` might be any string/object/identifier from a network endpoint, queue, SMS gateway, etc.

At the interface with EventFlow, we'll have two potential `Message` properties:

 * `"Received 123456abcdef"`
 * `"123456abcdef"`

If a user's monitoring this as a log, they're probably viewing lines of text showing the timestamp, level and `Message` property.

If the event scrolls by and looks like:

```
2017-02-03T12:34:56.123Z INF 123456abcdef
```

They won't have much context to determine what the event is supposed to mean, and probably wouldn't click in to see the `Message_1` property that would read `"Received 123456abcdef"`.

Switching the properties gets you:

```
2017-02-03T12:34:56.123Z INF Received 123456abcdef
```

which seems more useful at a glance.

It's a tough problem to solve without having some set of "reified" properties given special meaning by the framework. Maybe that would make sense for something as fundamental as a message or exception?

Another option (employed by one of the "flat" JSON formats output from Serilog) is to prefix reified properties with some kind of escape sequence, e.g. `$Message` or `@Message`; the set of acceptable properties still need to be laid out somewhere, but at least they can be extended without having to make API changes, and providers can apply some kind of sensible policy to handle reified properties that they don't make special use of. Just throwing ideas out there.

Cheers,
Nick

src/Microsoft.Diagnostics.EventFlow.Inputs.Serilog/LoggerSinkConfigurationEventFlowExtensions.cs
karolz-ms: I checked and it looks like we also do not throw in other places in EventFlow, just report an error through the health reporter. So your implementation is in fact very consistent

karolz-ms:
Thanks for a comprehensive reply, Nick, I learned something new today! Will probably make a similar change to the rest of the framework. Cheers!

PR #43, opened 2017-02-08
Add native ETW input

karolz-ms:
Also fix a few minor build warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ It runs in the same process as the application, so communication overhead is min
 - [PerformanceCounter](#performancecounter)
 - [Serilog](#serilog)
 - [Microsoft.Extensions.Logging](#microsoftextensionslogging)
+- [ETW (Event Tracing for Windows)](#etw-event-tracing-for-windows)
  
 **Outputs**
 - [StdOutput (console output)](#stdoutput)
@@ -142,8 +143,8 @@ This input listens to EventSource traces. EventSource classes can be created in 
 
 | Field | Values/Types | Required | Description |
 | :---- | :-------------- | :------: | :---------- |
-| `providerName` | provider name | Yes | This field specify what kind of input this is. For this input, it must be "EventSource" |
-| `level` | Critial, Error, Warning, Informational, Verbose, LogAlways | No | Specifies the collection trace level. Traces with equal or higher severity than specified are collected. For example, if Warning is specified, then Critial, Error, and Warning traces are collected. Default is All. |
+| `providerName` | provider name | Yes | Specifies the name of the EventSource to track. |
+| `level` | Critial, Error, Warning, Informational, Verbose, LogAlways | No | Specifies the collection trace level. Traces with equal or higher severity than specified are collected. For example, if Warning is specified, then Critial, Error, and Warning traces are collected. Default is LogAlways, which means "provider decides what events are raised", which usually results in all events being raised. |
 |`keywords` | An integer | No | A bitmask that specifies what events to collect. Only events with keyword matching the bitmask are collected, except if it's 0, which means everything is collected. Default is 0. |
 
 #### PerformanceCounter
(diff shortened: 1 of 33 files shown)


src/Microsoft.Diagnostics.EventFlow.EtwUtilities/project.json
```diff
@@ -0,0 +1,40 @@
+{
+  "version": "1.0.0-*",
+
+  "dependencies": {
+    "NETStandard.Library": "1.6.0"
```
brahmnes: Does it work if I want to run on netstandard 1.3?

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/EtwInput.cs
```diff
@@ -0,0 +1,163 @@
...
+        private EventFlowSubject<EventData> subject;
+        private ITraceEventSession session;
+        private bool isDisposed;
+
+        public EtwInput(IConfiguration configuration, IHealthReporter healthReporter)
+        {
+            Requires.NotNull(configuration, nameof(configuration));
+            Requires.NotNull(healthReporter, nameof(healthReporter));
+
+            IConfiguration providersConfiguration = configuration.GetSection("providers");
```
xiaomi7732: There are 2 places of the same string of 'providers' (in line 31 and line 34). I don't feel conformable seeing strings duplicated themselves in code. :-) Can we refactor that these strings are provided by EtwProviderConfiguration class as constants?

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/ITraceEventSession.cs
```diff
@@ -0,0 +1,21 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
```
xiaomi7732: nit: Remove unused usings.

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/StandardTraceEventSession.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            Requires.NotNull(healthReporter, nameof(healthReporter));
+
+            this.healthReporter = healthReporter;
+            this.inner = new TraceEventSession($"EventFlow-{nameof(EtwInput)}-{Guid.NewGuid().ToString()}", TraceEventSessionOptions.Create);
+            this.inner.Source.Dynamic.All += OnEvent;
+            this.isProcessing = false;
+        }
+
+        public event Action<EventData> AllEvents;
```
xiaomi7732: AllEvents serves a purpose that whenever the trace event happens, it got invoked. It is exposed as a public event here. At least one event handler is required, right? I feel it is a bit over exposed and it makes me feel exhausted to do null check every time is is used.
Can we make it a a private field as a simple delegate, that be passed in as a parameter in the constructor? We can then simply do null check once in the constructor and we don't need to worry about null reference in Process() or OnEvent().

src/Microsoft.Diagnostics.EventFlow.EtwUtilities/project.json
karolz-ms: Excellent thought, but unfortunately netstandard1.3 requires 4.6 and we depend on System.Diagnostics.Process which is not available for netstandard1.2 (which can do 451). Bottom line this is probably best we can do at the moment.

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
```diff
@@ -10,14 +10,12 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Diagnostics.EventFlow.Configuration;
 using Validation;
```
karolz-ms: will do

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
```diff
@@ -117,9 +115,9 @@ private void EnableInitialSources()
         private void EnableAsNecessary(EventSource eventSource)
         {
             // Special case: enable TPL activity flow for better tracing of nested activities.
-            if (eventSource.Guid == TplEventSourceGuid)
+            if (eventSource.Guid == TplActivities.TplEventSourceGuid)
             {
-                this.EnableEvents(eventSource, EventLevel.LogAlways, (EventKeywords)TaskFlowActivityIds);
+                this.EnableEvents(eventSource, EventLevel.LogAlways, (EventKeywords) TplActivities.TaskFlowActivityIdsKeyword);
```
karolz-ms: ok

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/EtwInput.cs
karolz-ms: sure

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/ITraceEventSession.cs
karolz-ms: done

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/StandardTraceEventSession.cs
karolz-ms: Very good suggestion. I ended up requiring a delegate to be passed to the Process() method of the ITraceEventSession interface.

README.md
```diff
@@ -292,6 +293,48 @@ namespace LoggerEventFlow
...
+| `type` | "ETW" | Yes | Specifies the input type. For this input, it must be "ETW". |
+| `providers` | JSON array | Yes | Specifies ETW providers to collect data from. |
+
+*Providers object*
+
+| Field | Values/Types | Required | Description |
+| :---- | :-------------- | :------: | :---------- |
+| `providerName` | provider name | Yes(*) | Specifies the name of the ETW provider to track. |
+| `providerGuid` | provider GUID | Yes(*) | Specifies the GUID of the ETW provider to track. |
+| `level` | Critial, Error, Warning, Informational, Verbose, LogAlways | No | Specifies the collection trace level. Traces with equal or higher severity than specified are collected. For example, if Warning is specified, then Critial, Error, and Warning traces are collected. Default is All. |
```
karolz-ms: Well, yes an no. Yes, LogAlways is 0 and Critical is 1. But LogAlways has a special meaning, which is "provider decides which events are raised" and by default it results in raising (and capturing) all events. 

I have updated the docs.

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/EtwInput.cs
```diff
@@ -0,0 +1,163 @@
...
+                    this.session.EnableProvider(providerConfiguration.ProviderGuid, providerConfiguration.Level, (ulong)providerConfiguration.Keywords);
+                }
+                else
+                {
+                    this.session.EnableProvider(providerConfiguration.ProviderName, providerConfiguration.Level, (ulong)providerConfiguration.Keywords);
+                }
+            }
+
+            // Also enable .NET Task Parallel Library hierarchical activity tracking
+            this.session.EnableProvider(TplActivities.TplEventSourceGuid, TraceEventLevel.Always, TplActivities.TaskFlowActivityIdsKeyword);
```
karolz-ms: Think about enabling this provider with this keyword as a flag that enables [hierarchical activity generation for .NET applications](https://blogs.msdn.microsoft.com/vancem/2015/09/14/exploring-eventsource-activity-correlation-and-causation-features/). This is low-cost thing, but maybe should be a "on by default, can turn off if necessary" thing. I will leave as-is for now.

PR #45, opened 2017-02-10
Allow filters to evaluate Enum values (like LogLevel)

FredrikGoransson:
Added value interpreter for Enum type properties (like LogLevel)

Filters can with this be expressed as:

"include": "Level == Informational"
"include": "Level != Error"
"include": "ProviderName == EventSourceName && Level == Error"
"include": "ProviderName == EventSourceName && Level <= Warning"
"include": "Level != Warning"

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/CachingPropertyExpressionEvaluator.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/CachingPropertyExpressionEvaluator.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Globalization;
+using System.Reflection;
 
 namespace Microsoft.Diagnostics.EventFlow.FilterEvaluators
 {
@@ -84,6 +85,16 @@ namespace Microsoft.Diagnostics.EventFlow.FilterEvaluators
                     return parsedDateTimeOffsetValue;
                 }
                 else return null;
+            }
+            else if (eventPropertyValueType.GetTypeInfo().IsEnum)
+            {
+                object retval = null;
+                try
+                {
+                    retval = Enum.Parse(eventPropertyValueType, this.value);
+                }
+                catch (Exception) { }
+                return retval;
             }
             else
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/CachingPropertyExpressionEvaluator.cs
```diff
@@ -85,6 +86,16 @@ private object InterpretRhsValue(Type eventPropertyValueType)
                 }
                 else return null;
             }
+            else if (eventPropertyValueType.GetTypeInfo().IsEnum)
+            {
+                object retval = null;
+                try
+                {
+                    retval = Enum.Parse(eventPropertyValueType, this.value);
```
karolz-ms: Enum.TryParse() is probably a bit more efficient

karolz-ms:
Thanks a bunch for your contribution, Fredrik! Much appreciated.

nblumhardt:
Is there a way to distinguish between the enum value `Error` and a property called `Error`?

E.g. assuming the following event from Serilog:

```csharp
var result = someApi.SomeCall();
if (result.HasError)
{
    // The event is an error, carrying a property called `Error` with an integer value
    Log.Error("The API call failed with code {Error}", result.ErrorCode);
}
```

Now if the filter is:

```
Level == Error
```

Is it possible the event will fail to match the filter, because `Error` is a property with value `123`? This would be a bit of a gotcha if so, since a monitoring system might not pick up a legitimate error event.

karolz-ms:
@nblumhardt great question. I do not think it is possible today because LHS of the expression is always interpreted as property name and RHS is always interpreted as a literal value. See https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/FilterParser.peg So there is no ambiguity.

nblumhardt:
@karolz-ms ah that makes sense, thanks for the clarification.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/CachingPropertyExpressionEvaluator.cs
FredrikGoransson: Did not use TryParse since both overloads are generic methods, eg TryParse<T> and we only have a Type at resolution time.

FredrikGoransson: md removed my < and >.... TryParse of T are the only overloads for TryParse

FredrikGoransson: md removed my < and >.... TryParse of T are the only overloads for TryParse

karolz-ms: ah, good point

PR #49, opened 2017-02-21
Fix nuget package name typo in README.md

dkultasev:
There is typo in the nuget package name "Microsoft.Diagnostic.EventFlow.Outputs.StdOutput" it is supposed to be "Microsoft.Diagnostic**s**.EventFlow.Outputs.StdOutput"


karolz-ms:
Thank you!

Issue #47, opened 2017-02-12
Configuration of eventFlowConfig.json in Service Fabric at Application level


JonJam:
I would like to be able to override settings in eventFlowConfig.json when using Service Fabric at the Application level.

This is order that I can have specify a different Application Insights Key in an Application Parameter file i.e. Dev, Pre-Prod and Prod and then have the Service use that key, rather than hard-coding it at the service level.

karolz-ms:
@JonJam thanks for the suggestion, it is a very good one. We should be able to add this capability in near future.

karolz-ms:
Will publish the new NuGet packages shortly

PR #50, opened 2017-02-27
Fix issue 47

karolz-ms:
This PR fixes issue https://github.com/Azure/diagnostics-eventflow/issues/47 by adding ability to refer to Service Fabric settings and application parameters from EventFlow configuration

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
@@ -4,8 +4,11 @@
 // ------------------------------------------------------------
 
 using System;
+using System.Diagnostics;
 using System.Fabric;
 using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.Diagnostics.EventFlow;
 using Microsoft.Extensions.Configuration;
 using Validation;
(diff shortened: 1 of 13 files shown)


src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
```diff
@@ -34,8 +40,53 @@ public static DiagnosticPipeline CreatePipeline(string healthEntityName, string
...
+                        }
+                        else
+                        {
+                            configurationRoot[kvp.Key] = newValue;
+                        }
+                    }
+                }
+                catch (RegexMatchTimeoutException)
+                {
+                    continue;
```
xiaomi7732: If the reg-match timed out, the exception will be swallowed. I am a bit concerned of that. I think we should at least give end user a hint on the exception. How about adding the exception message into the health report?

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
```diff
@@ -14,6 +17,9 @@ namespace Microsoft.Diagnostics.EventFlow.ServiceFabric
 {
     public static class ServiceFabricDiagnosticPipelineFactory
     {
+        public static readonly string ConfigurationPackageName = "Config";
```
xiaomi7732: nit: Is this intended to be the same as ServiceFabricConfigurationProvider.DefaultConfigurationPackageName? I can't determine that buy reading the code. If it is, I think it would be good to reuse the constant just in case it changes for some unforeseen reason. 
That said, not very worried. I don't have a problem keep it as is either.

karolz-ms: That is good suggestion. I decided to use the definition in ServiceFabricConfigurationProvider

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
karolz-ms: Makes sense, added. It is highly unlikely that the regex we use here will cause a timeout, even if the value is really long, but let's just make sure we are prepared for it.

PR #64, opened 2017-03-24
Include MessageTemplate in events from the Serilog input

### LINKED ISSUE #63: Template missing from Serilog input
The template in Serilog messages is a great way to consistently match the source of the message independent of the values in the message. Optionally including the template as part of the output would be very useful.

nblumhardt:
The `MessageTemplate` is used in lieu of an explicit "event type" by Serilog and many of its sinks. This PR attaches this property to events from the input.

Fixes #63.


karolz-ms:
Fantastic, thank you!

Issue #48, opened 2017-02-14
What additional standard metadata items should be included in EventFlow?


karolz-ms:
The primary reason for attaching metadata to events is to enable outputs to produce various semantically-rich telemetry types. In general, metadata has no fixed schema, but for the sake of standardization and to ease output implementation EventFlow has the notion of "standard metadata". The standard metadata defines property names and types associated with well-known metadata tags. Currently two kinds of standard metadata are defined: "metric" and "request": https://github.com/Azure/diagnostics-eventflow#standard-metadata-types

We are planning to add some more standard metadata types and would appreciate feedback on what additional standard metadata types should be added to EventFlow, either to its core, or as a well-known metadata for one of its outputs?

For example, Microsoft Application Insights has notions of "dependency call", "event" (representing a significant user interaction or significant configuration change), "exception" (unexpected error) and "page view". The dependency call seems like a useful and generic concept that could be added to the core, but the others I am less sure about.

nblumhardt:
In addition to the ones you've noted, I think `Timestamp` etc. also come under this heading (though the way they're represented may be different) - from EventData.cs:

```csharp
        public DateTimeOffset Timestamp { get; set; }
        public string ProviderName { get; set; }
        public LogLevel Level { get; set; }
        public long Keywords { get; set; }
```

Some observations based on the fields we used in Serilog and how this has mapped to the various back-ends and integrations:

 * `Timestamp` and `Level` concide with first-class metadata properties in Serilog 👍
 * `Message` is core to the representation of most log data; it'd be hard to reliably implement logging-oriented outputs without it
 * `Exception` has proven useful for some back-ends and for custom rendering - it's often given special treatment on the collection-side
 * Some notion of an event type is pretty common, but the semantics vary substantially
  - Serilog treats `MessageTemplate` as a surrogate event type (events logged with the same template have the same type), effectively globally distinct
  - MEL has both `EventId` (string or int), distinct per logger, and message templates (not always present, but global when it is)
  - NLog's about to get message template support
  - ETW etc. (as you'd have observed) has a pretty strong event id concept, distinct by source
 * `Source` appears almost universally in logging libraries - Serilog has a well-known, though not first-class property called `SourceContext` for this

The event type one is really powerful, but also hard to handle semantically. Not sure if there's a clean way to get end-to-end fidelity there.

FWIW, there's an interesting food-for-thought list in: https://github.com/nlog/nlog/wiki/Layout-Renderers.

[...]

PR #81, opened 2017-04-28
Add new metadata types

karolz-ms:
1. Add request and exception standard metadata
2. Add "event" metadata (ApplicationInsights-specific)
3. Add support for request and exception metadata to ApplicationInsights output and Elasticsearch output

This resolves issue https://github.com/Azure/diagnostics-eventflow/issues/48

Documentation update will be handled as a separate change

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/DataRetrievalResult.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/DataRetrievalResult.cs
@@ -3,6 +3,7 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+using System;
 using Validation;
 
 namespace Microsoft.Diagnostics.EventFlow.Metadata
@@ -12,23 +13,19 @@ namespace Microsoft.Diagnostics.EventFlow.Metadata
         Success = 0,
         MetadataPropertyMissing = 1,
         InvalidMetadataValue = 2,
-        DataMissingOrInvalid = 3
+        DataMissingOrInvalid = 3,
+        InvalidMetadataType = 4
     }
 
     public class DataRetrievalResult
     {
-        private static readonly DataRetrievalResult SuccessResult = new DataRetrievalResult() { Status = DataRetrievalStatus.Success };
+        public static readonly DataRetrievalResult Success = new DataRetrievalResult() { Status = DataRetrievalStatus.Success };
 
         public DataRetrievalStatus Status { get; private set; }
         public string Message { get; private set; }
 
         // Prevent public construction
         private DataRetrievalResult() { }
-
-        public static DataRetrievalResult Success()
-        {
-            return SuccessResult;
-        }
 
         public static DataRetrievalResult MissingMetadataProperty(string missingPropertyName)
         {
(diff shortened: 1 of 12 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/EventMetadata.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Collections.Generic;
 using Validation;
+using System.Diagnostics;
```
xiaomi7732: nit: Sort usings, please :-)

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/EventMetadata.cs
```diff
@@ -57,5 +58,39 @@ public override int GetHashCode()
...
+                return DataRetrievalResult.DataMissingOrInvalid(eventPropertyName);
+            }
+            else
+            {
+                eventPropertyValue = value;
+            }
+
+            return DataRetrievalResult.Success;
+        }
+
```
xiaomi7732: nit: Remove empty lines

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/DataRetrievalResult.cs
```diff
@@ -12,24 +13,20 @@ public enum DataRetrievalStatus
...
-        private static readonly DataRetrievalResult SuccessResult = new DataRetrievalResult() { Status = DataRetrievalStatus.Success };
+        public static readonly DataRetrievalResult Success = new DataRetrievalResult() { Status = DataRetrievalStatus.Success };
 
         public DataRetrievalStatus Status { get; private set; }
         public string Message { get; private set; }
 
         // Prevent public construction
         private DataRetrievalResult() { }
 
-        public static DataRetrievalResult Success()
```
xiaomi7732: This is surly simpler. But is this change only to make it simpler? The reason I ask here is that the other DataRetrievalResult are all methods, and this is becoming a public field, I am afraid that is a bit confusion from the caller perspective. What do you think? Or is there some other intention that I don't see?

src/Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsights/ApplicationInsightsOutput.cs
```diff
@@ -6,17 +6,20 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+
```
xiaomi7732: nit: Sort usings, please :-)

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/EventMetadata.cs
karolz-ms: ok

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/EventMetadata.cs
karolz-ms: done

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/Metadata/DataRetrievalResult.cs
karolz-ms: For the Success property there is simply no good reason to use a method. Other static methods in this class are factory methods that need parameters. I do not think this is particularly confusing.

src/Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsights/ApplicationInsightsOutput.cs
karolz-ms: In this file the usings are grouped and sorted within the group.  The groups are system namespaces, non-system library namespaces and EventFlow namespaces.

karolz-ms:
Thanks, Saar, for a quick review! Appreciate it!!

xiaomi7732:
@karolz-ms Thanks for adding new supported types.

Issue #77, opened 2017-04-20
Support for encrypted Service Fabric settings


zyrow123:
extend the support for Service Fabric settings and application parameters to allow for encrypted settings as well.

zyrow123:
Just noticed this comment in the code.
// We omit encrypted values due to security concerns--if you need them, use Service Fabric APIs to access them

So it is intentionally committed, so if I wanted to encrypt my instrumentation key for application insights and then still use the ServiceFabricDiagnosticPipelineFactory how would I go about doing that

karolz-ms:
Honestly there is no good way...

Maybe we are too restrictive, let me think about it

karolz-ms:
Resolved with Service Fabric NuGet package version 1.1.3

PR #83, opened 2017-05-08
Better Service Fabric configuration support

karolz-ms:
Adds support for Service Fabric encrypted settings
EventFlow configuration can be placed in arbitrary configuration package

Resolves issue https://github.com/Azure/diagnostics-eventflow/issues/77

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ConfigurationProvider/SecureStringExtensions.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ConfigurationProvider/SecureStringExtensions.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Runtime.InteropServices;
 using System.Security;
+using Validation;
 
 namespace Microsoft.Extensions.Configuration.ServiceFabric
 {
@@ -13,10 +14,7 @@ namespace Microsoft.Extensions.Configuration.ServiceFabric
     {
         public static string ToUnsecureString(this SecureString secureString)
         {
-            if (secureString == null)
-            {
-                throw new ArgumentNullException();
-            }
+            Requires.NotNull(secureString, nameof(secureString));
 
             IntPtr unmanagedString = IntPtr.Zero;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ConfigurationProvider/SecureStringExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace Microsoft.Extensions.Configuration.ServiceFabric
+{
+    public static class SecureStringExtensions
+    {
+        public static string ToUnsecureString(this SecureString secureString)
+        {
+            if (secureString == null)
```
xiaomi7732: nit: Is there reason not using the function from Validate library?

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
```diff
@@ -21,16 +21,32 @@ public static class ServiceFabricDiagnosticPipelineFactory
         public static readonly string FabricConfigurationValueReference = @"servicefabric:/(?<section>\w+)/(?<name>\w+)";
         public static readonly string FabricConfigurationFileReference = @"servicefabricfile:/(?<filename>.+)";
 
-        public static DiagnosticPipeline CreatePipeline(string healthEntityName, string configurationFileName = "eventFlowConfig.json")
+        public static DiagnosticPipeline CreatePipeline(
+            string healthEntityName, 
+            string configurationFileName = "eventFlowConfig.json",
```
xiaomi7732: nit: I know this is not related to this change but can we make this a constant as well since we are touching this signature anyway. No big deal. Your call.

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ConfigurationProvider/SecureStringExtensions.cs
karolz-ms: Fixed. This code was copied from somewhere else and I did not pay enough attention, good catch!

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
karolz-ms: Makes sense, fixed

karolz-ms:
Cool, thanks for the review, Saar!

PR #84, opened 2017-05-11
Expressions can use hexadecimal values

karolz-ms:


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/CachingPropertyExpressionEvaluator.cs
```diff
@@ -86,14 +88,32 @@ private object InterpretRhsValue(Type eventPropertyValueType)
                 }
                 else return null;
             }
+            // Special-case long, ulong, int and uint to allow hexadecimal notation
+            else if (eventPropertyValueType == typeof(long))
+            {
```
yantang-msft: Add byte and int16

karolz-ms: Makes sense, added.

karolz-ms:
Thanks for the review, Yanming!

Issue #80, opened 2017-04-24
Need an operator for the "is keyword set" test


karolz-ms:
In the current implementation none of the operators supported by expression allow for easy testing of individual keywords. We need an operator for that (to test that all bits in the keyword mask are set)

Something like this:
```json
  "filters" : [
    {
      "type": "drop",
      "include": "Keywords &== 0x4"
    }
  ]
```

We probably also need an operator that checks whether none of the bits in the keyword mask are set (perhaps !== makes sense as a symbol for this operation)

gperrego:
Yes Yes Yes Please.

yantang-msft:
#85

karolz-ms:
Core package version 1.1.4 has the operator &== implemented. 

Feel free to reactivate this issue if the !== operator is also necessary in your scenario.

PR #85, opened 2017-05-12
Add a new operator '&==' to address issue #80

yantang-msft:
Add a new operator '&=='. This is used to test the bit mask of the property value, which can be used to filter on properties like Keywords.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/BitwiseEqualityEvaluator.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/BitwiseEqualityEvaluator.cs
@@ -1,6 +1,7 @@
-﻿//-----------------------------------------------------------------------------
-// Copyright (c) Microsoft Corporation.  All rights reserved.
-//-----------------------------------------------------------------------------
+﻿// ------------------------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------------------------------------------
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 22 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/BitwiseEqualityEvaluator.cs
```diff
@@ -0,0 +1,73 @@
+﻿//-----------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
```
karolz-ms: This should be the header indicating MIT license, see other files in the repo

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/BitwiseEqualityEvaluator.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+            else
+            {
+                isValidRhsValue = ulong.TryParse(value, out this.unsignedRhsValue)
+                    && long.TryParse(value, out this.signedRhsValue);
+            }
+
+            if (!isValidRhsValue)
+            {
+                throw new ArgumentException($"The rhs value '{value}' of &= operator is not integral type");
```
karolz-ms: "right-hand side value" will be more understandable

also the operator is &==, not just &=

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/FilterEvaluators/BitwiseEqualityEvaluator.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public override bool Evaluate(EventData e)
+        {
+            object propertyValue;
+            if (!e.TryGetPropertyValue(this.propertyName, out propertyValue))
+            {
+                return false;
+            }
+
+            var valueAsString = propertyValue.ToString();
```
karolz-ms: Serializing to string and parsing back will be expensive from performance standpoint. Can we do better?

karolz-ms: We already have boxed the value so "is int", "is long" etc. should be really low cost. Yes, it is not as concise as your proposed implementation, but much faster!

PR #86, opened 2017-05-22
EventFlow pipeline simplification and hardening

karolz-ms:
This change addressed some outstanding issues with EventFlow pipeline

1. Pipeline disposal is too complex and at the same time will take longer than necessary if throttling occurred during operation. This is because we employ event counting scheme to determine when the pipeline is drained, but if events are throttled, the count will be off and we will wait till completion timeout for events that were throttled long ago.

2. Because of how BroadcastBlock works we might not warn the consumer of the pipeline if throttling occurs.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
@@ -1,0 +1,215 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+using Validation;
+
+namespace Microsoft.Diagnostics.EventFlow
+{
+    /// <summary>
+    /// The purpose of this class is to ensure that EventFlow pipeline does not drop events without warning through the health reporter.
+    /// This is a possiblity with BroadcastBlock, which, by design, will discard the current message once a new message arrives and the 
+    /// current message has been offered to all downstream blocks. This happens regardless whether the message was accepted, postponed
+    /// or discarded by the downstream block. 
+    /// The LossReportingPropagatorBlock works around this issue by reporting a message loss when a passing message is discarded by the downstream block,
+    /// or when a message is postponed and then not available for consumption later on.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal class LossReportingPropagatorBlock<T> : IPropagatorBlock<T, T>
+    {
+        private IHealthReporter healthReporter;
+        private ITargetBlock<T> target;
+        private ISourceBlock<T> source;
+        private bool propagateCompletion;
+        private bool completed;
+        private Lazy<TaskCompletionSource<bool>> completionSource;
+        private long postponements;
+        private long consumptionAttempts;
+
+        public LossReportingPropagatorBlock(IHealthReporter healthReporter)
+        {
+            Requires.NotNull(healthReporter, nameof(healthReporter));
+            this.healthReporter = healthReporter;
+            this.target = null;
+            this.propagateCompletion = false;
+            this.completed = false;
+            this.completionSource = new Lazy<TaskCompletionSource<bool>>(LazyThreadSafetyMode.PublicationOnly);
+            this.postponements = this.consumptionAttempts = 0;
+        }
+
+        public Task Completion { get { return this.completionSource.Value.Task; } }
+
+        public void Complete()
+        {
+            this.completed = true;
+            this.completionSource.Value.TrySetResult(true);
+
+            if (this.propagateCompletion)
+            {
+                Debug.Assert(this.target != null);
+                this.target.Complete();                
+            }
+
+            if (this.postponements > this.consumptionAttempts)
+            {
+                // Once we are completed, it won't be possible to consume any postponed messages and any postponed messages will effectively be lost.
+                this.healthReporter.ReportThrottling();
+            }
+        }
+
+        public void Fault(Exception exception)
+        {
+            this.completed = true;
+
+            if (this.propagateCompletion)
+            {
+                Debug.Assert(this.target != null);
+                this.target.Fault(exception);
+
+                // If the pipeline is faulted, the data loss is pretty much inevitable, so it does not make that much sense to report data loss here.
+                // (we theoretically could check for this.postponements > this.consumptionAttempts).
+            }
+            else
+            {
+                this.completionSource.Value.TrySetException(exception);
+            }
+        }
+
+        public IDisposable LinkTo(ITargetBlock<T> target, DataflowLinkOptions linkOptions)
+        {
+            Requires.NotNull(target, nameof(target));
+            if (this.target != null)
+            {
+                throw new InvalidOperationException($"{nameof(LossReportingPropagatorBlock<T>)} only supports a single target");
+            }
+
+            this.target = target;
+            if (linkOptions != null)
+            {
+                // Note DataflowLinkOptions.Append does not matter given that we are working with a single target only 
+                // and we will not enforce DataflowLinkOptions.MaxMessages
+                this.propagateCompletion = linkOptions.PropagateCompletion;
+            }
+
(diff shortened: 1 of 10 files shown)


test/TestHelpers/TaskUtils.cs
```diff
@@ -0,0 +1,29 @@
...
+// ------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.EventFlow.TestHelpers
+{
+    public static class TaskUtils
+    {
+        private static readonly TimeSpan JustABit = TimeSpan.FromMilliseconds(20);
```
yantang-msft: Can use SpinWait.SpinUtil directly instead of writing this helper method

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
```diff
@@ -0,0 +1,209 @@
...
+            this.propagateCompletion = false;
+            this.completed = false;
+            this.completionSource = new Lazy<TaskCompletionSource<bool>>(LazyThreadSafetyMode.PublicationOnly);
+        }
+
+        public Task Completion
+        {
+            get
+            {
+                if (this.propagateCompletion)
```
yantang-msft: Any reason passing the CompletionTask from the target block when propagateCompletion? Remember the bufferBlock will complete regardless of the subsequent's completion task. I think it's true for other blocks. Although it doesn't matter right now, it's better to be consistent. And we can get rid of the this.completed and only use one tcs.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
```diff
@@ -0,0 +1,209 @@
...
+            Verify.Operation(this.source != null, "The block must have a source");
+            Verify.Operation(object.ReferenceEquals(this.target, target), "Only single target is supported");
+        }
+
+        private void VerifySourceInitiatedOperation(ISourceBlock<T> source)
+        {
+            Verify.Operation(this.target != null, "The block must have a target");
+            // Make sure we set the source reference only once
+            Interlocked.CompareExchange(ref this.source, source, null);
+            Verify.Operation(object.ReferenceEquals(this.source, source), "Only single source block is supported");
```
yantang-msft: Assert source is BroadCastBlock

yantang-msft: for once

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
```diff
@@ -0,0 +1,209 @@
...
+                    break;
+
+                case DataflowMessageStatus.Declined:
+                case DataflowMessageStatus.DecliningPermanently:
+                case DataflowMessageStatus.NotAvailable:
+                    this.healthReporter.ReportThrottling();
+                    break;
+
+                case DataflowMessageStatus.Postponed:
+                    // This is a bit tricky--the message might be successfully consumed later, or it might no longer be available.
```
yantang-msft: Could add more comment here explain what the trick is, like BroadCastBlock will store one message, so it might successfully get the postponed message later on. But if new message arrives at the BroastCaseBlock and overrides the message, then it will fail at ConsumeMessage.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
```diff
@@ -0,0 +1,209 @@
...
+            switch(targetResponse)
+            {
+                case DataflowMessageStatus.Accepted:
+                    // Great, nothing to do
+                    break;
+
+                case DataflowMessageStatus.Declined:
+                case DataflowMessageStatus.DecliningPermanently:
+                case DataflowMessageStatus.NotAvailable:
+                    this.healthReporter.ReportThrottling();
```
yantang-msft: I feel like the switch is not needed here and just pass the call to the target. Whether the events will be dropped can determined by the return value BroadCastBlock's ConsumeMessage solely. To be more specific
1) NotAvailable will be a subsequent result of a failed ConsumeMessage call, so this will be duplicated with the ReportThrottling in ConsumeMessage.
2) It won't return DecliningPermanently, that happens when the target block is completed, when that happens, the current block is also completed.
3) It won't return Declined. Since there is always a source (unlike the first BufferBlock in the pipeline), it will return Postponed if the message is bounded by capacity.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            if (this.completed)
+            {
+                return false;
+            }
+
+            bool reserved = this.source.ReserveMessage(messageHeader, this);
+            if (!reserved)
+            {
+                // If we cannot reserve the message, it means it is gone before our target had a chance to consume it.
```
yantang-msft: This is not true. Source can return false if the message is already reserved by some other targets.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
```diff
@@ -0,0 +1,209 @@
...
+            {
+                // Message is no longer available
+                this.healthReporter.ReportThrottling();
+                return default(T);
+            }
+        }
+
+        public void ReleaseReservation(DataflowMessageHeader messageHeader, ITargetBlock<T> target)
+        {
+            VerifyTargetInitiatedOperation(target);
```
yantang-msft: It feels like this Reserve/Release APIs will and should never be used on this class.

test/TestHelpers/TaskUtils.cs
karolz-ms: Thanks for pointing out this method to me, I did not realize we had it.

But no, I do not think this is a perfect replacement, primarily because the SpinWait.SpinUntil() does not use Task Parallel Lib at all. With my custom thread scheduler used by the tests I want the scheduler to have a chance to do "something else" while we are waiting on condition.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
karolz-ms: You are right. Fixed

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
karolz-ms: Well, it does not _have to_ be...

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
karolz-ms: That is essentially what the comment says already, but I clarified it a bit. BTW the logic here is valid for any kind of block, not just BroadcastBlock, in the sense that postponement does not guarantee successful consumption (reservation, i.e. success of ReserveMessage(), does).

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
karolz-ms: My aim with this implementation was to create something that makes sense for variety of situations, not just in the context of being a downstream block for BroadcastBlock. It also makes it easier to verify if the block understands the full DataFlow protocol with all the possible responses handled.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
karolz-ms: Thx, I changed the comment to add this case too

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LossReportingPropagatorBlock.cs
karolz-ms: See my earlier comment about general applicability of this block.

Issue #87, opened 2017-05-22
System.ArgumentException: The key already existed in the dictionary


timomta:
Hello,

I am using the Application Insights output and the EventSource input with an application that I wrote for Service Fabric. My application itself is heathy, but the EventSource pipeline is throwing this exception (copied from the Service Fabric Explorer window):

Error event: SourceId='WeatherAPI-WeatherAPIService-DiagnosticsPipeline', Property='Connectivity'.
Diagnostics data upload has failed.
System.ArgumentException: The key already existed in the dictionary.
   at System.Collections.Concurrent.ConcurrentDictionary`2.System.Collections.Generic.IDictionary.Add(TKey key, TValue value)
   at Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsightsOutput.AddProperties(ISupportProperties item, EventData e, Boolean setTimestampFromEventData)
   at Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsightsOutput.SendEventsAsync(IReadOnlyCollection`1 events, Int64 transmissionSequenceNumber, CancellationToken cancellationToken) 

What is this error telling me? I'm not sure what dictionary it is keeping, its keys, or what that has to do with connectivity.

karolz-ms:
@timomta What is the version of the EventFlow.Core package that your application is using?

timomta:
Hello,

Microsoft.Diagnostics.EventFlow.Core 1.1.4

karolz-ms:
@timomta I think I know where the problem might be. Do any of your EventSource methods (event methods) have parameters named "keywords", "level" or "providerName" by any chance?

timomta:
I am using it to capture the events sent by Service Fabric's ServiceEventSource.cs, ServiceEventSource class. If you create a Service Fabric project in visual studio, you'll see this file/class. It does have a static class named Keywords:
        // Event keywords can be used to categorize events. 
        // Each keyword is a bit flag. A single event can be associated with multiple keywords (via EventAttribute.Keywords property).
        // Keywords must be defined as a public class named 'Keywords' inside EventSource that uses them.
        public static class Keywords
        {
            public const EventKeywords Requests = (EventKeywords)0x1L;
            public const EventKeywords ServiceInitialization = (EventKeywords)0x2L;
        }

There are also Event methods that pass in level:

        private const int MessageEventId = 1;
        [Event(MessageEventId, Level = EventLevel.Informational, Message = "{0}")]
[...]

PR #88, opened 2017-05-23
Eliminate the possibility of property name conflict for AI output

karolz-ms:


karolz-ms:
Related to issue https://github.com/Azure/diagnostics-eventflow/issues/87

PR #94, opened 2017-06-07
Inputs.Etw : Keep the original TraceEvent Timestamp

CITnDev:


karolz-ms:
@CITnDev  ETW package 1.1.2 has been published. Thanks again!

Issue #93, opened 2017-06-06
The ability to add Nested Scopes when using MicrosoftLogging input


JoeL1862:
Would it be possible to add the ability to add all nested scopes in the payload.

`using (logger.BeginScope("Test:{Test}", "Test"))
                    {
logger.LogInformation("Test Scoped Info");
                        using (logger.BeginScope("Test1:{Test1}", "Test1"))
                        {
                            logger.LogInformation("Test Scoped 2 Info");
                        }
                        logger.LogInformation("Test Scoped Info");
                    }`

Here it would be nice for the payload to have a property of "Test" with value "Test" for log entries before and after the nested scope. Then the log entry in the nested scope would have a payload with properties, "Test" = "Test" and "Test1" = "Test1".

karolz-ms:
This is great suggestion, thanks. We'll look into it!

karolz-ms:
Resolved by https://www.nuget.org/packages/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/1.1.1 or newer

PR #100, opened 2017-06-15
Support nested scopes for MicrosoftLogging input

yantang-msft:
Address issue #93 by having a scope that pointing to its parent scope.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
@@ -24,6 +24,8 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
         public EventFlowLoggerScope Parent { get; private set; }
 
 #if NET451
+        // LogicalCallContext will flow through cross app domain calls.
+        // Thus we make the FieldKey domain specific and wrap the EventFlowLoggerScope in ObjectHandle to avoid exceptions when cross app domain call happens.
         private static readonly string FieldKey = $"{typeof(EventFlowLoggerScope).FullName}_{AppDomain.CurrentDomain.Id}";
         public static EventFlowLoggerScope Current
         {
@@ -31,7 +33,7 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
             {
                 ObjectHandle handle = (ObjectHandle)CallContext.LogicalGetData(FieldKey);
 
-                // Unwrap the scope if it was set in the same AppDomain (as FieldKey is AppDomain-specific). 
+                // Unwrap the scope if it was set in the same AppDomain (as FieldKey is AppDomain-specific).
                 if (handle != null)
                 {
                     return (EventFlowLoggerScope)handle.Unwrap();
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
@@ -5,6 +5,7 @@
 
 using System;
 #if NET451
+using System.Runtime.Remoting;
 using System.Runtime.Remoting.Messaging;
 #else
 using System.Threading;
@@ -23,16 +24,24 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
         public EventFlowLoggerScope Parent { get; private set; }
 
 #if NET451
-        private static readonly string FieldKey = $"{typeof(EventFlowLoggerScope).FullName}";
+        private static readonly string FieldKey = $"{typeof(EventFlowLoggerScope).FullName}_{AppDomain.CurrentDomain.Id}";
         public static EventFlowLoggerScope Current
         {
             get
             {
-                return (EventFlowLoggerScope)CallContext.LogicalGetData(FieldKey);
+                ObjectHandle handle = (ObjectHandle)CallContext.LogicalGetData(FieldKey);
+
+                // Unwrap the scope if it was set in the same AppDomain (as FieldKey is AppDomain-specific). 
+                if (handle != null)
+                {
+                    return (EventFlowLoggerScope)handle.Unwrap();
+                }
+
+                return null;
             }
             private set
             {
-                CallContext.LogicalSetData(FieldKey, value);
+                CallContext.LogicalSetData(FieldKey, new ObjectHandle(value));
             }
         }
 #else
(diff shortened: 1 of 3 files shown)


karolz-ms:
Pull request # 100... must be good one! :-)

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
```diff
@@ -0,0 +1,75 @@
...
+        internal EventFlowLoggerScope(object state)
+        {
+            State = state;
+        }
+
+        public Object State { get; private set; }
+        public EventFlowLoggerScope Parent { get; private set; }
+
+#if NET451
+        private static readonly string FieldKey = $"{typeof(EventFlowLoggerScope).FullName}";
```
karolz-ms: Isn't FullName a string? Do we need to interpolate here?

lmolkova: this code is error prone for RPC calls.
1. EventFlowLoggerScope is not Serializable and does not inherit from MarshalRefByObject, which means that once cross appdomain call is done, app will crash with some SerializationException.
2. Even if you implement one or another way, second appdomain may have no knowledge about event flow - that is a typical case for app running in IIS. So when the second appdomain will initialize, app will crash not being able to resolve EventFlowLoggerScope 
3. Additionally, there are some issues with mutability of the object, potential copy during serialization, etc. so cross appdomain support is really complex.

So I suggest following (basically prevent things to be reused in the other app-domains):
1. Make FieldKey app-domain specific - add AppDomain.Current.Id to it
2. EventFlowLoggerScope wrap EventFlowLoggerScope  into ObjectHandle before string in the Logical call context.

This way, it will be successfully serialized, other appdomain will be happy to find ObjectHandle in the logical call context (will resolve the type), however no one will ever attempt to access EventFlowLoggerScope  from the other app-domain, so no actual resolution of EventFlowLoggerScope type, access and modification of object will happen.

This is the lesson learned from the Activity in the hard way :)

karolz-ms: Excellent advice, Liudmila.

Yanming, please put the gist of what L said in comments

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
```diff
@@ -0,0 +1,75 @@
...
+        public override string ToString()
+        {
+            return State?.ToString();
+        }
+
+        private class DisposableScope : IDisposable
+        {
+            public void Dispose()
+            {
+                Current = Current.Parent;
```
karolz-ms: I am wondering if there is a race condition here? I.e. two threads read the same value of Current.Parent

Maybe a lock on a private static object owned by DisposableScope would prevent that.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
lmolkova: @karolz-ms I don't remember the details, but do you really have to support NET 4.5.1? 
ILogger is declared in Microsoft.Extensions.Logging.Abstractions that targets netstandard1.1 only (which is compatible with 4.5, but you have to install whole .NET core stack to use it). 
May be I would be easier just to gracefully ignore scopes on 4.5 or do not store it in the logical call context there?

karolz-ms: @lmolkova @yantang-msft I would be fine with not supporting nested scopes with 4.5.x

karolz-ms: @lmolkova In that case I assume the AsyncLocal-based implementation does not have the concerns mentioned above?

lmolkova: yes, AsyncLocal does not flow into other AppDomains and does not have any of the above problems

yantang-msft: No need.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
yantang-msft: I'm not sure I understand your concern. It's AsyncLocal, so "Current" is different object for two threads.

yantang-msft: I actually have a test case for parallel scenario, unless you're talking about some other cases.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
yantang-msft: I followed what @lmolkova did for the Activity class.
Since we build provide Net451 and NetStandard1.1 packages, for the users who targets Net46, it will use the bits from Net451 package but not NetStandard1.1.
Considering the user base in full framework, I'd like to support the nested scope for 4.5.x as well.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
karolz-ms: Good point. AsyncLocal does not necessarily imply thread affinity, but it will be duplicated as necessary to maintain isolation. Disregard my comment.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLoggerScope.cs
```diff
@@ -0,0 +1,84 @@
...
+        public Object State { get; private set; }
+        public EventFlowLoggerScope Parent { get; private set; }
+
+#if NET451
+        private static readonly string FieldKey = $"{typeof(EventFlowLoggerScope).FullName}_{AppDomain.CurrentDomain.Id}";
+        public static EventFlowLoggerScope Current
+        {
+            get
+            {
+                ObjectHandle handle = (ObjectHandle)CallContext.LogicalGetData(FieldKey);
```
karolz-ms: Can you put in a comment explaining why we are using ObjectHandle here?

karolz-ms: Can you put in a comment explaining why we are using ObjectHandle here and not just storing the the EventFlowLoggerScope directly?

yantang-msft: Added.

PR #110, opened 2017-07-12
Nested scopes in MicrosoftLogging input logged as single Scope property

mfmadsen:
Current version of MicrosoftLogging input logs nested scopes to multiple Scope properties resulting in health check warnings as multiple property keys are not allowed. Also Scope values are assigned to different random keys (Scope, Scope_<x>).
This PR changes the behavior to log all nested scopes to a single Scope property with a value syntax:
OuterScope > InnerScope > InnerMostScope
The change causes the Scope property value to be stringified.
Tests have been modified to reflect the changes.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
@@ -7,7 +7,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Text;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Internal;
 
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
@@ -6,6 +6,8 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
+using System.Text;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Internal;
 
@@ -47,6 +49,8 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
             }
 
             var scope = EventFlowLoggerScope.Current;
+            List<string> scopePropertyValues = new List<string>(5);
+
             while (scope != null)
             {
                 if (scope?.State != null)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
```diff
@@ -61,17 +64,22 @@ public void Log<TState>(Extensions.Logging.LogLevel logLevel, EventId eventId, T
...
+                        scopePropertyValues.Add(scope.State.ToString());
                     }
                 }
 
                 scope = scope.Parent;
             }
 
+            if (scopePropertyValues.Count > 0)
+            {
+                AddPayloadProperty(properties, "Scope", string.Join(" > ", scopePropertyValues.AsEnumerable().Reverse()));
```
karolz-ms: 1. Could we use a stack, instead of a list, to avoid the Reverse() call?
2. I wonder if a better separator would be \r\n? I am afraid things might get a little confusing if any of the scope strings contain ">"

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
```diff
@@ -47,6 +48,8 @@ public void Log<TState>(Extensions.Logging.LogLevel logLevel, EventId eventId, T
             }
 
             var scope = EventFlowLoggerScope.Current;
+            List<string> scopePropertyValues = new List<string>(5);
```
karolz-ms: EventFlow today is not very good at avoiding allocations, but let's try to make things as lean as necessary. So

1. Let's use a pool of lists (or stacks, if you follow my other suggestion) and take a stack from the pool, or create it as necessary. You can use `ConcurrentBag<T>` for that, see https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs#L268 for example how to do that.
2. Avoid using a scope value stack at all if there is no scope.

mfmadsen:
Great suggestions. Will look into it and update.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
karolz-ms: I meant "as lean as possible"

mfmadsen: Both should now be handled in the newest update.

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
mfmadsen: Both should now be handled in the newest update.

mfmadsen:
Thanks!
Do I need to do anything else to complete this PR?
What is the expected time frame for the PR making it into a new nuget?

karolz-ms:
@mfmadsen Should be available within next couple of hours (look for package version 1.1.2)

PR #117, opened 2017-07-20
Add ReflectInsight output

calloncampbell:
We've added support for ReflectInsight as an output. Please take a look at what we've done and provide any feedback.

Thanks
@calloncampbell and @codematrix

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ It runs in the same process as the application, so communication overhead is min
 - [Azure EventHub](#event-hub)
 - [Elasticsearch](#elasticsearch)
 - [OMS (Operations Management Suite)](#oms-operations-management-suite)
+- [ReflectInsight](#reflectinsight)
 
 The EventFlow suite supports .NET applications and .NET Core applications. It allows diagnostic data to be collected and transferred for applications running in these Azure environments:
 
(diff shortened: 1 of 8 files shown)


karolz-ms:
Hi @calloncampbell  and @codematrix! Thanks a bunch for writing an output for ReflectInsight!

Before we proceed further I was just curious whether you have considered hosting the code for your output in a separate repository? I am thinking this might be ultimately better better for both Reflect and Microsoft--you'd have more control over how fast you want to evolve your output, what dependencies to take etc. Projects like Logstash, FluentD and Serilog are using this model very successfully, e.g. see https://www.fluentd.org/plugins

If you decide to go that route, I am totally fine with adding a reference to ReflectInsight output to EventFlow docs.

Let me know what you think and thanks again
Karol

codematrix:
Hi Karol,

Thanks for the feedback. I agree it would be best to have our own repo as suggested.  Having said that. Can we still keep the same Assembly Name and public key? For example, can we still have this in the **DiagnosticPipelineFactory.cs** file:

outputFactories["ReflectInsight"] = "**Microsoft.Diagnostics.EventFlow.Outputs.ReflectInsightOutputFactory, Microsoft.Diagnostics.EventFlow.Outputs.ReflectInsight**, Culture=neutral, PublicKeyToken=**b03f5f7f11d50a3a**";

This would assume we use your PublicKey.snk file and keep with same namespace format as shown above.

Please advise let us know if this is OK. If so, then we can provide another PR with only the README and DiagnosticPipelineFactory.cs change request.

Thanks,
Ross

karolz-ms:
Hi Ross, yes, we can have an entry in the DiagnosticPipelineFactory for your output. You can use any namespace you want, the proposed one is fine.

Just to clarify, even if the DiagnosticPipelineFactory entry is not there, you can still load your output from configuration by declaring the factory in the extensions section like described here: https://github.com/Azure/diagnostics-eventflow#creating-a-pipeline-with-custom-elements-using-configuration

Also, I do not think you want to use our public key, because you do not have the corresponding private key to sign your package (or at least you should not have, he he)

codematrix:
Thanks Karol. Sure we'll create our own key for signing. 
Once we are done will we can update the README and update the PipelineFactory file accordingly via PR. 

Thanks for all your help 😀
Ross

karolz-ms:
Sounds good! Closing this one for now.

Issue #124, opened 2017-08-07
EventSource hierarchical activity ids are missing unless another listner is also listening.


ghelyar:
I have a custom EventSource, which has Start and Stop events.

If I am collecting on that EventSource in PerfView or in Visual Studio 2017's "Diagnostic Events" window, I see activity IDs in the other viewer on both my Start and Stop events, and on my Start events only in EventFlow, with the Stop events offset by one level.

If I am only listening with EventFlow, as it would be in production, I do not see any hierarchical activity IDs (e.g. "//1/2/3") at all.

I got the same behaviour by using EventListener (with the TPL provider) directly.

Here is the output I see on .net 4.6.2 when using the EventSource input and a custom output that writes the payload serialised as JSON:

Without any other viewer:
```
{"EventId":1,"EventName":"OuterStart"}
{"EventId":3,"EventName":"InnerStart"}
{"EventId":4,"EventName":"InnerStop"}
{"EventId":2,"EventName":"OuterStop"}
```

With no code changes, but visual studio -> view -> other windows -> diagnostic events and the event source name added:
```
{"EventId":1,"EventName":"OuterStart","ActivityId":"//1/2/"}
{"EventId":3,"EventName":"InnerStart","ActivityId":"//1/2/1/","RelatedActivityId":"00000012-0000-0000-0000-0000bf999d59"}
{"EventId":4,"EventName":"InnerStop","ActivityId":"//1/2/"}
{"EventId":2,"EventName":"OuterStop"}
```

Notes:

- Without another viewer listening to the event source at the same time, activity ID is missing.
- Even when present, the stop events have the wrong activity IDs (when compared to either PerfView or VS' Diagnostic Events window)
https://stackoverflow.com/questions/42957473/what-could-be-the-reason-for-my-eventlistener-to-get-empty-activityid-values also shows this
The expected Activity IDs and Related Activity IDs (as shown by the other viewers at the same time) respectively would be
```
OuterStart: //1/2, null
InnerStart: //1/2/1, //1/2
InnerStop: //1/2/1, null
OuterStop: //1/2, null
```
- Related activity ID is not converted to an activity path. I can see from [EventDataExtensions.cs](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs#L41) that  ToString() is called rather than GetActivityPathString() for RelatedActivityId

[...]

PR #126, opened 2017-08-17
Fix deadlock in EventSourceInput

karolz-ms:
The problem was originally reported by users of ApplicationInsights EventSourceTelemetryModule, but I believe EventSourceInput suffers from the same problem.

Also fixed a small issue related to issue https://github.com/Azure/diagnostics-eventflow/issues/124 (RelatedActivityId should be path-decoded, just like ActivityID).

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Collections.Concurrent;
 using System.Diagnostics.Tracing;
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.Diagnostics.EventFlow.Configuration;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
```diff
@@ -57,6 +58,7 @@ public EventSourceInput(IReadOnlyCollection<EventSourceConfiguration> eventSourc
         }
 
         public IEnumerable<EventSourceConfiguration> EventSources { get; private set; }
+        public Task Initialization { get; private set; }
```
lmolkova: any reason (except tests) to keep it public?

karolz-ms: The main reason is somebody might really care about a particular "service initialization" event to be captured and this is not guaranteed until the initialization has been completed. But your question prompted me to seek a better way and indeed there is: we have an interface, `IRequiresActivation` that can be used for the pipeline to wait for initialization tasks to complete. I will change the implementation to use that.

Issue #134, opened 2017-09-14
Timestamp precision in EventSource input


ghelyar:
The [EventSource input](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs#L27) uses DateTime.UtcNow presumably because EventWrittenEventArgs does not contain a timestamp.

Would it be possible to use a more precise timestamp as described [here](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408%28v=vs.85%29.aspx)? The timestamps from the ETW input already use this level of precision, but of course they aren't added by EventFlow.

```
[DllImport("Kernel32.dll", CallingConvention = CallingConvention.Winapi)]
private static extern void GetSystemTimePreciseAsFileTime(out long filetime);

GetSystemTimePreciseAsFileTime(out long filetime);
var utcNow = DateTime.FromFileTimeUtc(filetime);
```

This requires interop and has minimum supported versions of Windows (8 and Server 2012), but perhaps it could be a configuration option or test whether it will work in a static constructor then fall back to DateTime.UtcNow if it won't, or something like that, or are these constraints too severe?

ghelyar:
We're just trying to capture events as accurately as possible for analysis. DateTime.UtcNow is not really accurate enough to get decent durations out of start/stop events, for example. Some events appear too short or even instant, while others appear too long.

karolz-ms:
Thanks, this helps. Would it then be sufficient if we used hi-res timer for just the full framework, and expect people to target Core 2.0 if they need hi-res timer for their Core apps?

ghelyar:
I'm using net framework anyway because service fabric can't do net core yet (except for hosting it as a guest executable) so that's enough for me.

I would have thought that saying target net core 2.0 would be sufficient for net core users. There's probably no reason to implement it for < 2.0 unless someone asks for it and has a reason they can't target 2.0.

karolz-ms:
Should be resolved with EventSource input version 1.1.3 or newer

Sorry it took so long to publish--we had major problems with our code signing infrastructure due to recent hurricanes

PR #137, opened 2017-09-27
Increase timestamping precision for EventSource input

karolz-ms:
This is to fix issue https://github.com/Azure/diagnostics-eventflow/issues/134

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs
@@ -28,6 +28,10 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
         {
             Debug.Assert(healthReporter != null);
 
+            // High-precision event timestamping is availabe on .NET 4.6+ and .NET Core 2.0+ 
+            // For the latter the implementation of DateTime.UtcNow has changed and we do not need to do anything.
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventDataExtensions.cs
```diff
@@ -7,24 +7,37 @@
...
+        [DllImport("Kernel32.dll", CallingConvention = CallingConvention.Winapi)]
+        private static extern void GetSystemTimePreciseAsFileTime(out long filetime);
+#endif
+
         public static EventData ToEventData(this EventWrittenEventArgs eventSourceEvent, IHealthReporter healthReporter, string context)
         {
             Debug.Assert(healthReporter != null);
 
+#if NETSTANDARD1_6
+            DateTime now = DateTime.UtcNow;
```
lmolkova: JFYI: netcoreapp 2.0 clr has precise timestamp implemented with `GetSystemTimePreciseAsFileTime`
But netcoreapp 1.0, 1.1  would still have an imprecise timestamp.

I think it should not be solved (i.e. users are encouraged to use 2.0), but it could make sense to add comment about that

karolz-ms: Good point, will do, thanks!

Issue #140, opened 2017-10-06
Perpetual output using EventFlow + ApplicationInsights + NLog


ibebbs:
## Description

I wish to output events written to an [EventSource](https://msdn.microsoft.com/en-us/library/system.diagnostics.tracing.eventsource.aspx) to both StdOut and ApplicationInsights. As such I used EventFlow to listen to events from the EventSource and write the output to both destinations (I had previously tried using [ApplicationInsights.EventSourceListener](https://www.nuget.org/packages/Microsoft.ApplicationInsights.EventSourceListener/) but, for reasons yet to be determined, it simply would not capture any output to ApplicationInsights). Anyway, this seemed to work fine and output is indeed written to both destinations as expected.

Now I also wish to add NLog output to ApplicationInsights so installed the [ApplicationInsights.NLogTarget](https://www.nuget.org/packages/Microsoft.ApplicationInsights.NLogTarget/). Again, this seemed to work as expected and messages written to NLog were indeed sent to ApplicationInsights.

Unfortunately, in the above scenario, if you delay writing to NLog long enough for libraries to initialize themselves fully, when the first NLog message is written you will see an unending stream of data containing event output from the "Microsoft-ApplicationInsights-Data" event source and find that the program enters a perpetual loop of writing data to the EventFlow output (including - most damagingly - the ApplicationInsights output).

## Reproduction

A full reproduction of the issue, with steps, can be found in this [repo](https://github.com/ibebbs/AiNlogEventSource)

## Investigation

By cloning EventFlow and building a debug version of Microsoft.Diagnostics.EventFlow.Inputs.EventSource (and temporarily disabling the strong name check), I was able to break at the pertinent points in the EventSourceInput.cs file. The following was noted:

1. A call to ```OnEventSourceCreated``` was made for the 'Microsoft-ApplicationInsights-Data' event source and passed to the 'EnableAsNecessary' method. As the 'Microsoft-ApplicationInsights-Data' was not included in the EventFlow config (and not a 'special case') no call to ```EnableEvents``` was made.
2. A subsequent call to ```OnEventWritten``` is (incorrectly) made for an event from the 'Microsoft-ApplicationInsights-Data' event source and, as it's not a 'special case', passed to the output subject and, ultimately, the output modules.

This can be seen in the call stack here:

![Call stack](https://raw.githubusercontent.com/ibebbs/AiNlogEventSource/master/Images/Call%20Stack.png)

[...]

PR #144, opened 2017-10-20
Add capability to disable EventSources

karolz-ms:
A useful feature in itself, but also constitutes a workaround for [issue 140](https://github.com/Azure/diagnostics-eventflow/issues/140)

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,20 +143,31 @@ This input listens to EventSource traces. EventSource classes can be created in 
 }
 ```
 
-*Top Object*
+*Top object*
 
 | Field | Values/Types | Required | Description |
 | :---- | :-------------- | :------: | :---------- |
 | `type` | "EventSource" | Yes | Specifies the input type. For this input, it must be "EventSource". |
 | `sources` | JSON array | Yes | Specifies the EventSource objects to collect. |
 
-*Sources Object*
+*Source object (element of the sources array)*
 
 | Field | Values/Types | Required | Description |
 | :---- | :-------------- | :------: | :---------- |
-| `providerName` | provider name | Yes | Specifies the name of the EventSource to track. |
-| `level` | Critial, Error, Warning, Informational, Verbose, LogAlways | No | Specifies the collection trace level. Traces with equal or higher severity than specified are collected. For example, if Warning is specified, then Critial, Error, and Warning traces are collected. Default is LogAlways, which means "provider decides what events are raised", which usually results in all events being raised. |
-|`keywords` | An integer | No | A bitmask that specifies what events to collect. Only events with keyword matching the bitmask are collected, except if it's 0, which means everything is collected. Default is 0. |
+| `providerName` | EventSource name | Yes(*) | Specifies the name of the EventSource to track. |
+| `providerNamePrefix` | EventSource name prefix | Yes(*) | Specifies the name prefix of EventSource(s) to track. For example, if the value is "Microsoft-ServiceFabric", all EventSources that have names starting with Microsoft-ServiceFabric (Microsoft-ServiceFabric-Services, Microsoft-ServiceFabric-Actors and so on) will be tracked. |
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -143,20 +143,31 @@ This input listens to EventSource traces. EventSource classes can be created in
...
+|`keywords` | An integer | No(**) | A bitmask that specifies what events to collect. Only events with keyword matching the bitmask are collected, except if it's 0, which means everything is collected. Default is 0. |
+
+*Remarks*
+
+(*) Out of `providerName`, `providerNamePrefix` and `disabledProviderNamePrefix`, only one can be used for a single source. In other words, with a single source one can enable an EventSource by name, or enable a set of EventSources by prefix, or disable a set of EventSources by prefix.
+
+(**) `level` and `keywords` can be used for enabling EventSources, but not for disabling them. Disabling events using level and keywords is not supported (but one can use level and/or keywords to *selectively enable* a subset of events from a given EventSource).
+
+(***) There is an issue with .NET frameworks 4.6 and 4.7, and .NET Core framework 1.1 and 2.0 where dynamically created EventSource events are dispatched to all listeners, regardless whether listners subscribe to events from these EventSources; for more information see https://github.com/dotnet/coreclr/issues/14434  `disabledProviderNamePrefix` property can be usesd to suppress these events.<br/>
```
brahmnes: typo: "listners"

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/Configuration/EventSourceConfiguration.cs
```diff
@@ -12,13 +15,62 @@ namespace Microsoft.Diagnostics.EventFlow.Configuration
...
+
+            if (!string.IsNullOrWhiteSpace(ProviderName))
+            {
+                return eventSource.Name == ProviderName;
+            }
+            else if (!string.IsNullOrWhiteSpace(ProviderNamePrefix))
+            {
+                return eventSource.Name?.StartsWith(ProviderNamePrefix, StringComparison.Ordinal) ?? false;
+            }
+            else return false;
```
brahmnes: You don't need "else"

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
```diff
@@ -155,46 +158,95 @@ private void EnableAsNecessary(EventSource eventSource)
...
+        private void BroadcastEventIfSourceNotDisabled(EventWrittenEventArgs eventArgs)
+        {
+            string eventSourceName = eventArgs.EventSource.Name;
+            if (eventSourceName != null)
+            {
+                bool isDisabled;
+                if (!this.disabledSources.TryGetValue(eventSourceName, out isDisabled))
+                {
+                    isDisabled = this.EventSources.Any(eventSourceConfiguration => eventSourceConfiguration.Disables(eventArgs.EventSource));
+                    this.disabledSources[eventSourceName] = isDisabled;
```
brahmnes: Maybe I don't remember the architecture of eventflow. Instead of checking this on every event, can't it work like, the provider gets disabled up front, and you won't see its events at all?

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/Configuration/EventSourceConfiguration.cs
karolz-ms: I know :-)

src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs
karolz-ms: That is what the issue in the documentation is about--with this bug we are getting events even if we do not enable the provider (EventSource) at all. So we need to check where every events is coming from

brahmnes: Ah ok

PR #148, opened 2017-11-10
Exposed more ElasticSearch config to users

andersosthus:
Exposes the following settings to be user-defined:

- `NumberOfShards` (Default: 1)
- `NumberOfReplicas` (Default: 5)
- `RefreshInterval` (Default 15s)

Added a test to verify that the `ElasticSearchOutputConfiguration` is properly populated.
Updated `README.MD` to reflect those changes.

Fixed a type in `README.MD` regarding Index prefix. The example showed the Index prefix with a trailing `-`. This is not needed since it's added by code.

Updated the Date generation for Index name to use leading zeros (instead of 2017.5.22 it will be 2017.05.22). This is in-line with how all official ElasticSearch tools (Logstash, Beats) generate dates for time-based series.

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ This output writes data to the [Elasticsearch](https://www.elastic.co/products/e
 ```json
 {
     "type": "ElasticSearch",
-    "indexNamePrefix": "app1-",
+    "indexNamePrefix": "app1",
     "serviceUri": "https://myElasticSearchCluster:9200",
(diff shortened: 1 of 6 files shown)


andersosthus:
Great. I'll bump the version on monday morning :)

karolz-ms:
Version 2.2.0 of the ES output package has been pushed to NuGet

PR #153, opened 2017-11-21
Implement http plugin

mblaschke:
Sends event log informations to webserver in different formats:
- `Json` object with array items
- `JsonLines` with events each as own json object separated by newlines

Using HTTP as transport protocol


karolz-ms:
Thank you, Markus, for your contribution! Looks very interesting.

Just a caveat: there is a big holiday coming this week in the US (Thanksgiving). Everything slows down pretty much to a crawl. Just a heads up that we might be a bit slow regarding reviewing this PR and merging it in, but I will try my best.

karolz-ms:
Would be great to see some unit tests (e.g. that the output "obeys" its configuration settings and outputs the data in the expected format)

src/Microsoft.Diagnostics.EventFlow.Outputs.Http/Http.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                    default:
+                        break;
+                }
+
+                HttpContent contentPost = new StringContent(payload, Encoding.UTF8, configuration.HttpContentType);
+
+                var result = httpClient.PostAsync(new Uri(configuration.ServiceUri), contentPost).Result;
+
+                return Task.FromResult(true);
```
karolz-ms: I assume you wanted to check the result here and report problem as necessary?

src/Microsoft.Diagnostics.EventFlow.Outputs.Http/Http.cs
```diff
@@ -0,0 +1,171 @@
...
+                    break;
+
+                default:
+                    errorMessage = $"{nameof(configuration)}: unknown ContentType \"{nameof(this.configuration.ContentType)}\" specified";
+                    healthReporter.ReportWarning(errorMessage, EventFlowContextIdentifiers.Configuration);
+                    break;
+            }
+        }
+
+        public Task SendEventsAsync(IReadOnlyCollection<EventData> events, long transmissionSequenceNumber, CancellationToken cancellationToken)
```
karolz-ms: Make it an async method

SergeyKanzhelev:
What are the expected consumers of this data? 

Another qq - how well will it work with [Azure Time Series](https://azure.microsoft.com/services/time-series-insights/)?

src/Microsoft.Diagnostics.EventFlow.Outputs.Http/Http.cs
```diff
@@ -0,0 +1,171 @@
...
+                        payload = serializer.Serialize(new XmlWriterSettings { Indent = false }, events);
+                        break;
+
+                    default:
+                        break;
+                }
+
+                HttpContent contentPost = new StringContent(payload, Encoding.UTF8, configuration.HttpContentType);
+
+                var result = httpClient.PostAsync(new Uri(configuration.ServiceUri), contentPost).Result;
```
karolz-ms: await on the result

src/Microsoft.Diagnostics.EventFlow.Outputs.Http/Http.cs
```diff
@@ -0,0 +1,171 @@
...
+            try
+            {
+                string payload = "";
+
+                switch (configuration.ContentType) 
+                {
+                    case "text":
+                        foreach (EventData evt in events)
+                        {
+                            payload += evt.ToString() + "\n";
```
karolz-ms: Consider using StringBuilder class to assemble the payload

src/Microsoft.Diagnostics.EventFlow.Outputs.Http/Http.cs
```diff
@@ -0,0 +1,171 @@
...
+using Microsoft.Diagnostics.EventFlow.Configuration;
+using Newtonsoft.Json;
+using Validation;
+using ExtendedXmlSerializer.Configuration;
+using ExtendedXmlSerializer.ExtensionModel.Xml;
+
+namespace Microsoft.Diagnostics.EventFlow.Outputs
+{
+
+    public class Http : IOutput
```
karolz-ms: Let's call it "HttpOutput"--"Http" is too ambiguous

src/Microsoft.Diagnostics.EventFlow.Outputs.Http/Http.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                    case "json-lines":
+                        foreach (EventData evt in events)
+                        {
+                            payload += JsonConvert.SerializeObject(evt) + "\n";
+                        }
+                        break;
+
+                    case "xml":
+                        var serializer = new ConfigurationContainer().Create();
```
karolz-ms: (minor) Would it make sense to create the serializer during startup and just cache it in a class variable?

mblaschke:
I've removed XML because I'm not happy with the solution but JSON should be totally fine.
I've also implemented TCP and UDP streams because they nearly the same solution (but without HTTP overhead and currently without TLS). I also changed the `contentType` to `format` for better understanding.

For TCP I'm not 100% sure if I need the lock to ensure no mixed content when multiple threads are sending data.

Tests are still on TODO.

@SergeyKanzhelev
eg. logstash or whatever user wants.. for json you can attach every backend system 😃

mblaschke:
Oh i also had to change the TargetFramework for TCP/UDP to dotnet 2.0 because of `Dns.GetHostAddresses` and currently no idea if I find a better solution for dotnet 1.6

karolz-ms:
@mblashke I would prefer if you created a separate PR for TCP & UDP outputs. Smaller PRs are easier to review, but the main reason is the TCP/UDP output will have different sets of problems, since it is a much lower-level mechanism than HTTP.

Before you do that, can you explain what is the motivation for the TCP/UDP outputs? Have you considered gRPC as a substitute for raw TCP or UDP?

mblaschke:
@karolz-ms
I thought that it would maybe easier because the handling is quite the same but just different protocols.

Because TCP and UDP are just low level protocols the idea was to reduce the parsing and traffic overhead to a minimum in high load environments. Logstash for example supports HTTP, TCP and UDP collecting so this PR is the "logstash support feature".

karolz-ms:
@mblaschke It HTTP and TCP looks similar but the fact that you need to manage TCP connections yourself makes an important difference and increases complexity. For example, the way you lock on tcpClient it your current implementation may cause you to run out of threadpool threads if the destination is slow. Under "right" circumstances that will cause your service (the consumer of EventFlow) to stop making any progress. I am sure you can solve this, but it is more work and testing, and most likely not the only thing than needs improvement.

Also keep in mind that if you are after high throughput applications, the way you serialize data is at least as important as the connection protocol. You won't get very fast using uncompressed JSON. That is why I mentioned gRPC in my previous question, which by default uses efficient ProtoBuf serialization.

I appreciate your willingness to help, Markus, but with raw TCP/UDP I am hesitant to add such output to EventFlow (and advertise it as general-purpose component) unless someone comes to us and shows that there is a real-world need for it, and the implementation can be competitive, in terms of speed and reliability, with a custom output using gRPC. Perhaps publishing an example would be a better way to get people started.

mblaschke:
I wasn't fully sure about the lock and that's why I also mentioned it.
I just wanted to share the implementation because it's just straight forward and not complex to maintain and provides logstash support.
Also it's easy to debug because you only need to start netcat to just collect the readable traffic and the developer only has to change the configuration :)

You're right, protobuf would make more sense in a high load environment. The implementation can still be done as format but it's not as easy as json.

However I'm also fine with only the HTTP connection.

karolz-ms:
OK, thanks for understanding, so let's focus on the HTTP output

mblaschke:
@karolz-ms 
Finished :)

Warsaw.sln
```diff
@@ -35,44 +35,44 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Nuspecs", "Nuspecs", "{3B5E
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Utilities", "Utilities", "{C0F892AE-D5F5-42FD-943C-6428CA252650}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.EventFlow.Core", "src\Microsoft.Diagnostics.EventFlow.Core\Microsoft.Diagnostics.EventFlow.Core.csproj", "{F1EF768D-9E8B-42D5-B0E1-C7EF2270C245}"
```
karolz-ms: Unnecessary project type GUID change, please revert

Warsaw.sln
```diff
@@ -333,6 +335,30 @@ Global
...
 		{F7BB746D-9863-4831-B950-8FC64A39ACB6}.Release|x64.Build.0 = Release|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Debug|x64.Build.0 = Debug|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Release|Any CPU.Build.0 = Release|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Release|x64.ActiveCfg = Release|Any CPU
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341}.Release|x64.Build.0 = Release|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
karolz-ms: Remove this GUID (probably from deleted project)

Warsaw.sln
```diff
@@ -333,6 +335,30 @@ Global
...
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Debug|x64.Build.0 = Debug|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Release|Any CPU.Build.0 = Release|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Release|x64.ActiveCfg = Release|Any CPU
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452}.Release|x64.Build.0 = Release|Any CPU
+		{788653DF-CD08-4180-95EA-5349E26384CE}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
karolz-ms: This also looks like orphaned project GUID

Warsaw.sln
```diff
@@ -365,5 +391,8 @@ Global
 		{9BC644CD-01DC-4B3E-A6D0-B26AD07DE34F} = {D67068AF-16BB-4037-BE4E-B642419089BA}
 		{D8AD5E1D-BA05-4064-B3DF-2180CE680657} = {47200F40-43E1-4B09-B803-A921FED2BF05}
 		{F7BB746D-9863-4831-B950-8FC64A39ACB6} = {86A33E36-2266-4E31-8D91-1F720C9A1F27}
+		{B71FB566-9E87-4BE0-8E9D-12B59FB10341} = {5AD3BB0F-B6F8-4361-AC2D-CC3514DA37A0}
+		{CF9C8DD7-E627-46AC-A19D-BF9328BB3452} = {5AD3BB0F-B6F8-4361-AC2D-CC3514DA37A0}
```
karolz-ms: remove

Warsaw.sln
```diff
@@ -365,5 +391,8 @@ Global
...
+		{788653DF-CD08-4180-95EA-5349E26384CE} = {5AD3BB0F-B6F8-4361-AC2D-CC3514DA37A0}
```
karolz-ms: remove

Warsaw.sln
```diff
@@ -93,22 +93,24 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Diagnostics.Event
...
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights", "src\Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights\Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights.csproj", "{D8AD5E1D-BA05-4064-B3DF-2180CE680657}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights", "src\Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights\Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights.csproj", "{D8AD5E1D-BA05-4064-B3DF-2180CE680657}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestHelpers", "test\TestHelpers\TestHelpers.csproj", "{F7BB746D-9863-4831-B950-8FC64A39ACB6}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput", "src\Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput\Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput.csproj", "{B71FB566-9E87-4BE0-8E9D-12B59FB10341}"
```
karolz-ms: You probably want Core project type GUID (9A19...), not full framework GUID

src/Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput/Configuration/HttpOutputConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Diagnostics.EventFlow.Configuration
+{
+    // !!ACTION!!
+    // If you make any changes here, please update the README.md file to reflect the new configuration
+    public class HttpOutputConfiguration: ItemConfiguration
+    {
+        public static readonly string DefaultFormat = "json";
+
+        public string ServiceUri { get; set; }
+        public string Format { get; set; }
```
karolz-ms: Suggest defining an enumeration for this property and letting the configuration framework validate that a valid value is specified in the configuration file. E.g. see how TraceInput does this

src/Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput/HttpOutput.cs
```diff
@@ -0,0 +1,163 @@
...
+            try
+            {
+                var payload = new StringBuilder("");
+
+                switch (configuration.Format) 
+                {
+                    case "text":
+                        foreach (EventData evt in events)
+                        {
+                            payload.AppendLine(evt.ToString());
```
karolz-ms: Have you tested this?

EventData does not override ToString() so I do not think it will give you what you expect. It is going to produce "EventData object" or something similar for every event, no matter what data it carries.

I suggest for starters just dropping the text format and retaining json + json-lines

src/Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput/HttpOutput.cs
```diff
@@ -0,0 +1,163 @@
...
+                }
+
+                HttpContent contentPost = new StringContent(payload.ToString(), Encoding.UTF8, configuration.HttpContentType);
+
+                HttpResponseMessage response = await httpClient.PostAsync(new Uri(configuration.ServiceUri), contentPost);
+                response.EnsureSuccessStatusCode();
+            }
+            catch (Exception ex)
+            {
+                this.healthReporter.ReportProblem($"Fail to send events in batch. Error details: {ex.ToString()}");
```
karolz-ms: Add something to identify which output produced the error (e.g. nameof(HttpOutput)) and what URL was the data sent to.

src/Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput/HttpOutput.cs
```diff
@@ -0,0 +1,163 @@
...
+                throw;
```
karolz-ms: Do not throw here--just eat the exception. You have already reported the error so the exception is considered "handled" and there is nothing for the diagnostic pipeline to do beyond that.

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/HttpOutputTests.cs
```diff
@@ -0,0 +1,54 @@
...
+                var httpOutputConfiguration = new HttpOutputConfiguration();
+                configFragments[0].Bind(httpOutputConfiguration);
+
+                Assert.Equal(httpOutputConfiguration.ServiceUri, "http://localhost:1000");
+                Assert.Equal(httpOutputConfiguration.Format, "json-lines");
+                Assert.Equal(httpOutputConfiguration.BasicAuthenticationUserName, "mywebuser");
+                Assert.Equal(httpOutputConfiguration.BasicAuthenticationUserPassword, "mywebpass");
+                Assert.Equal(httpOutputConfiguration.HttpContentType, "application/x-custom");
+            }
+        }
```
karolz-ms: I'd like to see a test that verifies the output will produce different data (json vs json-lines), depending on the configuration option. 

This should not be too hard if you add HttpClient as a constructor parameter. In the output code you would create an instance if null (default) is passed. In the test code you would mock it using Moq

src/Microsoft.Diagnostics.EventFlow.Outputs.HttpOutput/HttpOutput.cs
mblaschke: I've used Stdout as example that's why the throw is also there

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/HttpOutputTests.cs
mblaschke: Hm.. I'm just a C# beginner and maybe I need some help with this goal.

mblaschke:
Can you update the Warsaw.sln? When i touch it with Visual Studio (for osx) it changes the whole file :(

karolz-ms:
OK, sorry for the delay, looking at getting this through right now

karolz-ms:
I pushed your changes into feature/http-output branch manually. Will do the remaining work there. I will open an issue to track the remaining work.

Thank you for the contribution, much appreciated!

PR #169, opened 2018-01-15
Add support for complex key in sf referenced configuration

ape-box:
The reason for this PR is that I would like to be able to reference Service Fabric configuration with keys containing symbols like dots, hyphens, and other common separators.

To support such key names, I loosened up the matching regex for the key name capture, as that is the part which more commonly contains structured keys as "Elasticsearch.Url", while the section part is not as frequently structured with separators.

This assumption may not be correct as it is based on my experience with Service Fabric, which is not very extensive, I apologize if mistaken.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.ServiceFabric/ServiceFabricDiagnosticPipelineFactory.cs
@@ -18,7 +18,9 @@ namespace Microsoft.Diagnostics.EventFlow.ServiceFabric
 {
     public static class ServiceFabricDiagnosticPipelineFactory
     {
-        public static readonly string FabricConfigurationValueReference = @"servicefabric:/(?<section>\w+)/(?<name>.+)";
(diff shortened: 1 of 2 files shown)


karolz-ms:
@ape-box thanks for your contribution!

The proposal makes a lot of sense in general and I agree that the current version of the regex is too restrictive. On the other hand allowing _anything_ in the name portion of the value reference is strikes me as perhaps too much of a good thing. How about allowing commonly used characters from a well-known set (in addition to the word characters already allowed)? For example we could allow
 
 - the dot (.), 
 - underscore (_), 
 - dash (-), 
 - colon (:), 
 - slash (/) 
 - and hash (#). 

I would expect this set to cover 99% of desired value names. What do you think?

ape-box:
Thank @karolz-ms for the prompt review, wrt regex I think that's better :)

I'm not sure what is the policy for comments, but i preferred to had a minimum of comment to facilitate reading in case someone is not that familiar with regex.

karolz-ms:
@ape-box FYI this was published on Nuget (version 1.1.5 of the EventFlow.ServiceFabric package)

ape-box:
Thank you

PR #174, opened 2018-01-26
Adding optional configuration switch to use Azure Gov base URL for OMS

MonasteryJohn:
I have a need to use EventFlow to send EventSource data from a Service Fabric application to Log Analytics in OMS. My specific OMS workspace lives in Azure's US Government Cloud. 

Based on the documentation [here](https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-services-monitoringandmanagement) the endpoints for the OMS API are slightly different for Gov cloud.

I added a new config setting for "UseAzureGov" as a boolean. This is used simply to switch the base API url from commercial to gov. No other changes are needed to make this work. I also updated the README file to include documentation on the new, optional, config setting.

I've ran all of the scripts requested in the guidelines. I've also built the Output.OMS DLL locally and tested that I do, in fact, get log entries in my gov OMS subscription.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Microsoft.Diagnostics.EventFlow
+﻿# Microsoft.Diagnostics.EventFlow
 
 ## Introduction
 The EventFlow library suite allows applications to define what diagnostics data to collect, and where they should be outputted to. Diagnostics data can be anything from performance counters to application traces.
@@ -668,7 +668,8 @@ The OMS output writes data to [Operations Management Suite](https://www.microsof
 {
   "type": "OmsOutput",
   "workspaceId": "<workspace-GUID>",
-  "workspaceKey": "<base-64-encoded workspace key>"
+  "workspaceKey": "<base-64-encoded workspace key>",
+  "useAzureGov" : "<optional-boolean>"
 }
 ```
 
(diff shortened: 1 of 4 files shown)


karolz-ms:
Hi @MonasteryJohn thank you for your contribution! Much appreciated, esp. the documentation update.

Before we merge, I have two suggestions
1. There is US government Azure flavor, but there is also Chinese government flavor and German government flavor. And it is not unlikely that in future there will be more. So instead of having a `useAzureGov` flag, how about having a parameter to override the OMS service domain (the `ods.opinsights.azure.us` part)? It could default to the public Azure (`ods.opinsights.azure.com`), but could be overridden as necessary for national clouds.
2. Please make sure that you bump up the minor version number in [the project file](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/Microsoft.Diagnostics.EventFlow.Outputs.Oms.csproj) (`VersionPrefix` setting). We need that to publish new version of the OMS output Nuget package.

Hope this makes sense. Thanks again!
Karol

MonasteryJohn:
Hi @karolz-ms thank you for your feedback. I changed the configuration setting to act as you requested and updated the VersionPrefix as well. Hopefully the way that I added it to this PR is acceptable, if not, please do let me know how to do it correctly.

I've not been able to get my changed file to deploy with my solution, so I've had to back off of it until I can get a properly signed one. Looking forward to getting this functionality in our application.

Thanks!

karolz-ms:
@MonasteryJohn this should be now on Nuget. Let us know if you run into any issues!

Issue #136, opened 2017-09-26
Set metadata metric name from event property


mikanyg:
The metric meta data does not support specifying metric name from an event property, only by specifying the metric name directly in json filter configuration.

I would like to be able to dynamically specify metric name based on an event property, just like in Request (requestNameProperty) and Dependency (targetProperty) meta data.

The scenario would be that we have Serilog input logger with pre-configured Context properties that is detected by the include filter. 

Usage in code could be 
`Log.Information("This message is converted to a metric {MetricName} with a count of {MetricValue}, "CustomMetricName", 1);`

example config of filter
{
  "type": "metadata",
  "metadata": "metric",
  "include": "ProviderName==Serilog && LogType==Metric",
  "metricNameProperty": "MetricName",
  "metricValueProperty": "MetricValue"
}

Which would make it easy to have a general metric meta data filter that can make logging configuration more maintainable.

karolz-ms:
Makes a ton of sense, thanks for the suggestion, Mikael. Will try to get that added. Feel free to submit a PR if you need it soon

mikanyg:
I don't expect to have time for a PR anytime soon, but looking forward to see it in action :-)

PR #176, opened 2018-01-30
Add support for metricNameProperty to metric metadata

shrimma:
Addresses issue #136.

Add support for setting metric name from an event property, similar to setting requests with 'requestNameProperty'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -812,12 +812,14 @@ Metrics are named time series of floating-point values. Metric metadata defines 
 | Field | Values/Types | Required | Description |
 | :---- | :-------------- | :------: | :---------- |
 | `metadata` | "metric" | Yes | Indicates metric metadata definition; must be "metric". |
-| `metricName` | string | Yes | The name of the metric. |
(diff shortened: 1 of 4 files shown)


yantang-msft:
@shrimma Thank you for your contribution. The change looks good.
Please also update the minor version of the project, so a new nuget package can be published.

shrimma:
@yantang-msft updated version number. Glad to contribute.

karolz-ms:
Thanks @shrimma and @yantang-msft !

I assume we still need to publish the new version of the Core assembly on Nuget? I can take care of that tomorrow, just confirm please--don't see it published yet

yantang-msft:
@karolz-ms Yes, we still need to publish the package.
I tried to publish it. However, our document is out of date and I don't have the latest password for the nuget account...

karolz-ms:
@yantang-msft @shrimma The core package with this change (version 1.1.11) is now published on Nuget. Thanks again!

PR #177, opened 2018-02-01
Update EventFlow.ServiceFabric to .NET standard 2.0

tomheijmans:
Because of the release of Microsoft.ServiceFabric 6.1.456 which now supports .NET standard 2.0 I updated the Nuget package Microsoft.Diagnostics.EventFlow.ServiceFabric to use it so it can reference .NET standard 2.0. Removed the nuspec file because it will now use the dotnet pack command.


nuspecs/ServiceFabric/content/PackageRoot/Config/eventFlowConfig.json
```diff
@@ -1,43 +0,0 @@
-{
-  "inputs": [
```
yantang-msft: @karolz-ms Should we keep this file so the provider name can be added automatically?

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/Microsoft.Diagnostics.EventFlow.ServiceFabric.csproj
```diff
@@ -5,14 +5,19 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <VersionPrefix>1.1.5</VersionPrefix>
```
yantang-msft: Upgrade the version so new packages can be published

src/Microsoft.Diagnostics.EventFlow.ServiceFabric/Microsoft.Diagnostics.EventFlow.ServiceFabric.csproj
```diff
@@ -5,14 +5,19 @@
...
+    <TargetFrameworks>netstandard2.0;net451</TargetFrameworks>
     <PlatformTarget>x64</PlatformTarget>
     <DelaySign>true</DelaySign>
     <AssemblyName>Microsoft.Diagnostics.EventFlow.ServiceFabric</AssemblyName>
     <AssemblyOriginatorKeyFile>../../PublicKey.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
-    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <PackageId>Microsoft.Diagnostics.EventFlow.ServiceFabric</PackageId>
+    <PackageTags>Microsoft, Diagnostics, EventFlow, Service Fabric</PackageTags>
```
yantang-msft: This is semicolon delimited

nuspecs/ServiceFabric/content/PackageRoot/Config/eventFlowConfig.json
tomheijmans: I removed this file, because the content-file is only added in projects which uses the packages.config. Most of the default ServiceFabric project templates uses .NET Core or ASP.NET Core, so they won't get the eventFlowConfig.json file added to their project. Now the behavior is the same in all kind of projects.

karolz-ms: I think that makes sense

tomheijmans:
@karolz-ms you're right on that. I've set the VersionPrefix  to 1.2.0 as you suggested.

karolz-ms:
@tomheijmans This should be now published to Nuget

tomheijmans:
@karolz-ms thank you!

Issue #182, opened 2018-02-16
Dependency conflict on Microsoft.Extensions.Logging 1.1.1 -> 2.0.0


eloekset:
I've got a Service Fabric project and an Azure Function sharing the same class libraries, and I've made a dependency on EventFlow for logging. It works fine for the Service Fabric project, but the Azure Function project crashes whenever the function is triggered with the following message:

*Method not found: 'Microsoft.Extensions.Logging.ILoggerFactory Microsoft.Diagnostics.EventFlow.Inputs.LoggerFactoryExtensions.AddEventFlow(Microsoft.Extensions.Logging.ILoggerFactory, Microsoft.Diagnostics.EventFlow.DiagnosticPipeline)'.*
![image](https://user-images.githubusercontent.com/791721/36314325-bedd64aa-1334-11e8-876b-2e350ef19022.png)

I've debugged this by creating a new Azure Function project with EventFlow, which worked fine until I added a dependency on **Microsoft.Extensions.Logging 2.0.0**.

The code that raises the exception is:
```
// Configure logging
var loggerFactory = new LoggerFactory()
      .AddEventFlow(diagnosticsPipeline);
```

In my real project I get an implicit dependency on 2.0.0, since I'm using **Microsoft.EntityFrameworkCore 2.0.1**. If I try to downgrade to **Microsoft.Extensions.Logging 1.1.1** I get the following error:
![image](https://user-images.githubusercontent.com/791721/36314251-8864cddc-1334-11e8-9e7b-6488adedef76.png)

I've published the sample app where I reproduced the error here: https://github.com/eloekset/EventFlowFunctionApp1/tree/master
And this branch has a dependency on **Microsoft.Extensions.Logging 1.1.1**: https://github.com/eloekset/EventFlowFunctionApp1/tree/feature/FixRuntimeError

I haven't yet figured out why the Service Fabric project works fine, when it also has the same dependencies on the class libraries using EFCore 2.0.1. The Service Fabric project hosts a service based on ASP.NET Core having target framework set to net47, and I've set the Function App project to net47 as well.

karolz-ms:
@eloekset can you try the workarounds specified in [.NET issue 481](https://github.com/dotnet/standard/issues/481), in particular, turning on auto binding redirect generation, and report if it helped?

eloekset:
[...]

PR #184, opened 2018-02-22
Build .NET Standard 2.0 assemblies

eloekset:
Ref issue #182, there's a dependency conflict between **Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging** and **Microsoft.Extensions.Logging** when building apps that reference .NET Standard 2.0 libraries. This gets very difficult to solve for an Azure Function app.

NuGet packages **Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging** and ****Microsoft.Diagnostics.EventFlow.Core** should contain libraries for .NET Standard 2.0 to fix the issue.

This PR can be a starting point. There's probably some work left to make it complete, but this is as far as I got.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/HealthReporters/CsvHealthReporter.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/HealthReporters/CsvHealthReporter.cs
@@ -506,6 +506,8 @@ namespace Microsoft.Diagnostics.EventFlow.HealthReporters
             }
 #elif NETSTANDARD1_6
             basePath = Path.GetDirectoryName(this.GetType().GetTypeInfo().Assembly.Location);
+#elif NETSTANDARD2_0
+            basePath = Path.GetDirectoryName(this.GetType().Assembly.Location);
 #endif
             this.Configuration.LogFileFolder = Path.Combine(basePath, this.Configuration.LogFileFolder);
         }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/HealthReporters/CsvHealthReporter.cs
```diff
@@ -506,6 +506,8 @@ private void ProcessLogFileFolder()
             }
 #elif NETSTANDARD1_6
             basePath = Path.GetDirectoryName(this.GetType().GetTypeInfo().Assembly.Location);
+#elif NETSTANDARD2_0
```
xiaomi7732: Instead of branch code here, I suggest we change the using in this file to:
```
#if NET451
using System.Web;
#else
using System.Reflection;
#endif
```
and unify the .NETStandard implementation to:
```
#else
basePath = Path.GetDirectoryName(this.GetType().GetTypeInfo().Assembly.Location);
#endif
```
It is more possible we have to target another new .NETStandard than another .NET Framework.

eloekset: Thanks! I've updated the PR.

karolz-ms:
@eloekset thank you for your contribution! I ended up creating [another PR](https://github.com/Azure/diagnostics-eventflow/pull/209) to cover a larger set of EventFlow assemblies, but your PR was a great starting point! I kept your commits so that you get the deserved credit for this improvement.

eloekset:
Thanks, @karolz-ms! I’m celebrating. 😁 https://twitter.com/eloekset/status/991424043634053122?s=21

PR #190, opened 2018-03-07
Add ability to monitor CLR events through ETW input

karolz-ms:


test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/EtwInputTests.cs
```diff
@@ -158,13 +163,72 @@ public void ReportsProblemsWhenConfigurationIsInvalid()
...
+                try
+                {
+                    throw new Exception("We seek perfection");
+                }
+                catch (Exception)  {  }
+
+                EventData exceptionEvent = null;
+                bool exceptionEventRaised = await TaskUtils.PollWaitAsync(() => observer.Data.TryDequeue(out exceptionEvent), TraceSessionActivationTimeout);
+                Assert.True(exceptionEventRaised);
+                Assert.Equal("We seek perfection", exceptionEvent.Payload["ExceptionMessage"]);
```
xiaomi7732: I like this expectation.

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/StandardTraceEventSession.cs
```diff
@@ -74,17 +75,23 @@ public void Process(Action<EventData> onEvent)
...
-                    // Suppress events from TplEventSource--they are mostly interesting for debugging task processing and interaction,
-                    // and not that useful for production tracing. However, TPL EventSource must be enabled to get hierarchical activity IDs.
-                    if (!TplActivities.TplEventSourceGuid.Equals(traceEvent.ProviderGuid))
-                    {
-                        onEvent(traceEvent.ToEventData(this.healthReporter));
-                    }
-                };
+                this.onEvent = onEvent;
+
+                this.inner.Source.Dynamic.All += TrackEvent;
```
xiaomi7732: Is it the case that the event handler will be accumulated if Process() is called multiple times? Looks like a potential memory leaking spot. Maybe remove the handler before adding it back?

karolz-ms: @xiaomi7732 that is not possible--`isProcessing` member variable is there to prevent that

Issue #180, opened 2018-02-08
Disable strong naming for local build to make unit tests easier to run


xiaomi7732:
We shall update the project that strong naming/signing only runs when it is the build intending for releasing or validation. It should be opt-in rather than opt-out.

karolz-ms:
https://github.com/Azure/diagnostics-eventflow/pull/196 solves this issue

xiaomi7732:
Thank you for taking care of this!

PR #196, opened 2018-04-06
Do not delay sign EventFlow assemblies

karolz-ms:
This change is intended to address [issue 180](https://github.com/Azure/diagnostics-eventflow/issues/180) I ended up with just removing the strong name signing altogether. It causes problems at dev time and on non-Windows platforms and I am not sure the cost is justified. Note that the assemblies in Nuget packages will still be signed as part of release build.


SergeyKanzhelev:
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/beaa444d435e806af7456862f9f3ba9668f6fcfe/Signing.targets#L41

SergeyKanzhelev:
https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/public-signing.md

karolz-ms:
Thanks, @SergeyKanzhelev! good to know 

After reading through the materials you shared I much more convinced that getting rid of strong name signing altogether is the right approach for EventFlow.

@xiaomi7732 a review would be appreciated :-)

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipelineFactory.cs
```diff
@@ -316,25 +316,25 @@ private static void ReportInvalidPipelineConfiguration(IHealthReporter healthRep
             Debug.Assert(healthReporter != null);
 
             inputFactories = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
-            inputFactories["EventSource"] = "Microsoft.Diagnostics.EventFlow.Inputs.EventSourceInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.EventSource, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a";
```
xiaomi7732: These looks like real-signed public token. I believe these are going to be used in the reflector for loading. Are these types still going to be loaded in signed bits in run-time?

src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/FilterParser.peg
```diff
@@ -1,6 +1,6 @@
 ﻿@namespace Microsoft.Diagnostics.EventFlow
 @classname FilterParser
-@accessibility internal
```
xiaomi7732: Why expose this as public now?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipelineFactory.cs
karolz-ms: Yes. The signing that remains is Authenticode signing that confirms the DLL was build by Microsoft; it does not affect the .NET assembly loader.

xiaomi7732: Okay. I don't think I understand the auth-code process but I'll trust your words on this. 💯

src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/FilterParser.peg
karolz-ms: Good catch, I think it was a leftover from some earlier experiments, but I tried again and as far as I can tell, this change is not necessary. I will revert it after pushing this PR.

PR #200, opened 2018-04-19
Delete the "suite" NuGet package

karolz-ms:
The "suite" NuGet package was meant to be an easy way to get started with EventFlow, but I do not think it adds much value--take a look at updated docs that assume it is not there and see for yourself--not that much difference. Also it does not quite work, and it never will, on .NET Core. I propose we get rid of it.

Also getting rid of private copy of several target files, which I believe are leftovers from MSBuild 14 times, although I am not 100% sure about that. Unless one tells me to retain them, I am going to just try deleting them and see if the rolling build complains.


README.md
```diff
@@ -31,55 +31,32 @@ The EventFlow suite supports .NET applications and .NET Core applications. It al
...
 
 ## Getting Started
-1. To quickly get started, you can create a simple console application in VisualStudio and install the Nuget package Microsoft.Diagnostics.EventFlow. 
-2. After the nuget package is installed, there should be a eventFlowConfig.json file added to the project. This file contains default configuration for the EventFlow pipeline. A few sections with optional configuration elements are commented outs. These sections can be enabled as desired. Here is what the file looks like
+1. To quickly get started, you can create a simple console application in VisualStudio and install the following Nuget packages:
+    * Microsoft.Diagnostics.EventFlow.Inputs.Trace
+    * Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsights
+    * Microsoft.Diagnostics.EventFlow.Outputs.StdOutput
+
+2. Add a JSON file named "eventFlowConfig.json" to your project and set the Build Action property of the file to "Copy if Newer". Set the content of the file to the following:
```
yantang-msft: You said it doesn't quite work, like what? The suite package doesn't work .net core, but in updated doc, the Trace input still doesn't work on .net core?

Installing 3 packages versus 1, plus add the config file, copy and paste, set build action to copy if newer, that's 5 more steps for the user. No matter how possible it is, there will be a higher chance of failure. Is there a strong reason behind that we want to remove this suite package?

karolz-ms: @yantang-msft newer versions of NuGet do not support adding anything other than binaries to the project. Neither the `eventFlowConfig.json` will be added, nor the included PowerShell script (that is supposed to ensure `eventFlowConfig.json` is copied to the output directory) will be executed. So the suite package does not add any value other than reducing the number of packages from 3 to 1.

yantang-msft: That's a huge breaking change! I can't find the doc, could you point me to it?

xiaomi7732:
NuGet package support has been developed. As a matter of fact, one can put content into the NuGet Packages now. Example below.

The suite package for now is 19k total download; where as the Core package has 100k download.
I think I am a bit leaning toward kill the package. A good example might be a better idea to bootstrap the usage.

Just FYI.

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
  <metadata>
    <!--Required metadata-->
    <id>$id$</id>
    // package properties ....................
    
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="AbcdefgPackage" version="$version$" exclude="Build,Analyzers" />
//         .... Other dependencies.
      </group>
    </dependencies>
    <contentFiles>
      <files include="any\netstandard2.0\eventFlowConfig.json" buildAction="None" copyToOutput="true" flatten="true" />
    </contentFiles>
  </metadata>

  <files>
    <!--config example-->
    <file src="..\Out\eventFlowConfig.json" target="contentFiles\any\netstandard2.0\eventFlowConfig.json"></file>
  </files>
</package>
```

karolz-ms:
So apparently things changes since I last looked at it, which was April last year. At that time I was told `contentFiles` are not supported for Core projects and there were no plans to add that support.

It looks like things have changes and support was added sometime after MSBuild 15 was rolled out (see [this blog](https://blog.nuget.org/20160126/nuget-contentFiles-demystified.html), "Supported Project Types" paragraph, October 2017 update. Swell.

Having said that I still don't think we want the "suite" package. In fact I think we should never have it shipped in the first place. Such container packages are a good idea when you have a lot of packages and there is high probability one will want to upgrade the whole set. ASP.NET Core is a good example. But in our case our input and output packages are not nearly as numerous and can be independently updated. So the container package does not provide any value other than marginally helping with the "getting started" scenario. And it is getting into one's way down the road--I had multiple questions from customers confused about whether they need this package or not, once they took explicit dependency on a newer version of an input or output package.

Issue #191, opened 2018-03-15
EventHub Output: How to output Event Hub Partition Key


tedvanderveen:
We use EventSource inputs and output events to EventHub using EventFlow. We want to process the events in Event Hub using partitioning to make sure related events are handled by the same Event Hub "consumer" instance.
So my question is, how can we provision an EventSource to output the PartitionKey that is outputted to Event Hub by EventFlow?

karolz-ms:
@tedvanderveen I assume you are asking about a mechanism to send event data to specific EventHub partition--is this correct?

If so, this is currently not supported, but could be relatively easily added

In your case, is it typical that a single service instance (or single EventFlow instance) writes to a single EventHub partition? Or is it the case that a single service instance will be writing to multiple EventHub partitions? I am asking because I am wondering if we can still batch events efficiently when explicit partitioning is used.

tedvanderveen:
@karolz-ms hi! We are writing to a single Event Hub that is configured with 8 partitions. Nu providing an EventHub PartitionKey, we can assure that related events are processed nu the same EventHub consumer instance.

karolz-ms:
@tedvanderveen thanks, I was thinking along same lines. So just to confirm, in your scenario, every EventSource event might potentially go to a different partition and would have to be examined individually?

tedvanderveen:
@karolz-ms every event on EventHub shall be processed individually, but we need to be able to ensure that related events are processed by the same EventHub consumer receiver instance. This is exactly what the EventHub PartionKey is meant for. At the moment the EventFlow output to EventHub functionality does not set this PartionKey system property in the output event message, while EventHub best practices documentation recommed using this EventHub event metadata field.

karolz-ms:
@tedvanderveen  We are not using Microsoft.ServiceBus.Messaging library; we are using Microsoft.Azure.EventHubs https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.eventhubs?view=azure-dotnet, which I believe is now the recommended library to interact with EH. 

In that library if you want to send events to specific partition, you use SendAsync overload that takes partitionKey parameter: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.eventhubs.eventhubclient.sendasync?view=azure-dotnet#Microsoft_Azure_EventHubs_EventHubClient_SendAsync_System_Collections_Generic_IEnumerable_Microsoft_Azure_EventHubs_EventData__System_String_

[...]

PR #203, opened 2018-04-23
Add optional support for Event Hub partitioned output

tedvanderveen:
See Issue #191: Event Hub has powerful partitioning features that can be leveraged when sending events with a partition key: [SendAsync(IEnumerable<EventData>, String)
](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.eventhubs.eventhubclient.sendasync?view=azure-dotnet#Microsoft_Azure_EventHubs_EventHubClient_SendAsync_System_Collections_Generic_IEnumerable_Microsoft_Azure_EventHubs_EventData__System_String)
This PR contains changes that allows for an optional Event Hub Output configuration property `partitionKeyProperty` that maps Event Data to the target `PartitionKey` value.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubFactory.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubFactory.cs
@@ -3,19 +3,30 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+using Microsoft.Diagnostics.EventFlow.Metadata;
+using Microsoft.Diagnostics.EventFlow.Outputs.EventHub;
 using Microsoft.Extensions.Configuration;
 using Validation;
 
 namespace Microsoft.Diagnostics.EventFlow.Outputs
 {
-    public class EventHubOutputFactory : IPipelineItemFactory<EventHubOutput>
+    public class EventHubOutputFactory : IPipelineItemFactory<IOutput>
     {
-        public EventHubOutput CreateItem(IConfiguration configuration, IHealthReporter healthReporter)
+        public IOutput CreateItem(IConfiguration configuration, IHealthReporter healthReporter)
         {
             Requires.NotNull(configuration, nameof(configuration));
             Requires.NotNull(healthReporter, nameof(healthReporter));
 
-            return new EventHubOutput(configuration, healthReporter);
+            //Check wether the optional partitionKeyProperty configuration element is set. If so, return an instance of PartitionedEventHubOutput
+            string partitioKeyProperty = configuration[PartitionKeyData.PartitionKeyPropertyMoniker];
+            if (string.IsNullOrWhiteSpace(partitioKeyProperty) == false)
+            {
+                return new PartitionedEventHubOutput(configuration, healthReporter);
+            }
+            else //return output handler that does not use Event Hub partinions
+            {
+                return new EventHubOutput(configuration, healthReporter);
+            }
         }
     }
 }
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Metadata/PartitionKeyData.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Metadata/PartitionKeyData.cs
@@ -1,0 +1,57 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+
+using Microsoft.Diagnostics.EventFlow.Metadata;
+using Validation;
+
+namespace Microsoft.Diagnostics.EventFlow.Outputs.EventHub
+{
+    /// <summary>
+    /// Object representing data associated with "PartitionKey" in Event Hub.
+    /// </summary>
+    internal class PartitionKeyData
+    {
+        internal static readonly string EventMetadataKind = "partitionKey";
+        internal static readonly string PartitionKeyPropertyMoniker = "partitionKeyProperty";
+
+        public string PartitionKey { get; private set; }
+
+        private PartitionKeyData()
+        {
+        }
+
+        public static DataRetrievalResult TryGetData(
+            EventData eventData,
+            EventMetadata eventTelemetryMetadata,
+            out PartitionKeyData partitionKeyData)
+        {
+            Requires.NotNull(eventData, nameof(eventData));
+            Requires.NotNull(eventTelemetryMetadata, nameof(eventTelemetryMetadata));
+            partitionKeyData = null;
+
+            if (!EventMetadataKind.Equals(eventTelemetryMetadata.MetadataType, System.StringComparison.OrdinalIgnoreCase))
+            {
+                return DataRetrievalResult.InvalidMetadataType(eventTelemetryMetadata.MetadataType, EventMetadataKind);
+            }
+
+            string partitionKey = eventTelemetryMetadata[PartitionKeyPropertyMoniker];
+            if (string.IsNullOrWhiteSpace(partitionKey))
+            {
+                return DataRetrievalResult.MissingMetadataProperty(PartitionKeyPropertyMoniker);
+            }
+
+            DataRetrievalResult result = eventTelemetryMetadata.GetEventPropertyValue(eventData, PartitionKeyPropertyMoniker, out partitionKey);
+            if (result.Status != DataRetrievalStatus.Success)
+            {
+                return result;
+            }
+
+            partitionKeyData = new PartitionKeyData();
+            partitionKeyData.PartitionKey = partitionKey;
+            return DataRetrievalResult.Success;
+        }
+    }
+}
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/PartitionedEventHubOutput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/PartitionedEventHubOutput.cs
@@ -119,7 +119,14 @@ namespace Microsoft.Diagnostics.EventFlow.Outputs
 
                     foreach (List<MessagingEventData> batch in batches)
                     {
-                        tasks.Add(hubClient.SendAsync(batch, partitionedEventData.Key));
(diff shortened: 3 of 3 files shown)


tedvanderveen:
@karolz-ms Yes you are right. I merged the new PartitionKey specific code changed into the existing `EventHubOutput` implementation. As you can see, most of the logic concerning the Event Hub batch size etc is still in place. No more code duplication.

tedvanderveen:
@karolz-ms if you have a minute, can you please setup the base for that unit test you are seeking to add?

karolz-ms:
@tedvanderveen sure, working on it..

karolz-ms:
@tedvanderveen  see https://github.com/Azure/diagnostics-eventflow/commit/3a04d785e7d80054334ed883881c3e401c4f24a1  Your test should follow along the same lines

tedvanderveen:
@karolz-ms I added tests for the Partitioning Event Hub output, all tests pass

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
```diff
@@ -83,52 +84,67 @@ public EventHubOutput(IConfiguration configuration, IHealthReporter healthReport
...
+                        if (batches.Count == 0 ||
+                            batchByteSize + messageSize > EventHubMessageSizeLimit)
+                        {
+                            batches.Add(new List<MessagingEventData>());
+                            batchByteSize = 0;
+                        }
+
+                        batchByteSize += messageSize;
+
+                        List<MessagingEventData> currentBatch = batches[batches.Count - 1];
```
karolz-ms: This is pre-existing issue, but can you move this assignment into the if statement body above, i.e. after line 110? We do not need to do this for every EventData instance, only when we create a new batch

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
```diff
@@ -168,10 +186,11 @@ void IDisposable.Dispose()
             IEventHubClient[] oldClients = Interlocked.Exchange<IEventHubClient[]>(ref this.clients, null);
 
             List<Task> tasks = new List<Task>();
-            foreach (EventHubClient client in oldClients)
+            foreach (IEventHubClient client in oldClients)
```
karolz-ms: Good catch, thanks!

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/EventHubOutputTests.cs
```diff
@@ -55,26 +55,68 @@ public void UsesIsoDateFormat()
...
             client.Verify(c => c.SendAsync(It.Is<IEnumerable<MessagingEventData>>(b => verifyBatch(b))), Times.Once);
             healthReporter.Verify(hr => hr.ReportWarning(It.IsAny<string>(), It.IsAny<string>()), Times.Never);
             healthReporter.Verify(hr => hr.ReportProblem(It.IsAny<string>(), It.IsAny<string>()), Times.Never);
         }
-    }
 
-    
-}
+        [Fact]
+        public async Task SendsDataToPartitionedEventHub()
```
karolz-ms: Cool. Can you please add one more test case that tests the following scenario: 

1. 3 EventData objects
2. One has PartitionByProperty == "A"
3. Second has PartitionByProperty == "B"
4. Third does not have a PartitionByProperty

.. and verify that expected IEventHubClient overloads are called with partition key set to the right values (we expect 3 different calls here).

After that is done it is good to go!

tedvanderveen: @karolz-ms I extended the tests to test these three cases (2 partitionKey values plus event without PartitionByProperty value). Also I added three events with big payload targeting same "Partiton2" value for PartitionByProperty, so also the batch splitting refactoring I applied is tested.

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
```diff
@@ -83,52 +84,51 @@ public EventHubOutput(IConfiguration configuration, IHealthReporter healthReport
...
-                List<List<MessagingEventData>> batches = new List<List<MessagingEventData>>();
-                int batchByteSize = 0;
+                List<Task> tasks = new List<Task>();
 
-                foreach (EventData eventData in events)
+                foreach (var partitionedEventData in groupedEventData)
                 {
-                    int messageSize;
-                    MessagingEventData messagingEventData = eventData.ToMessagingEventData(out messageSize);
+                    SendBatch(partitionedEventData.Select(e => new Tuple<MessagingEventData, int>(e.ToMessagingEventData(out var messageSize), messageSize)).ToList());
```
karolz-ms: If you use tuples, consider C# 7 build-in support https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7 that lets you name tuple members instead of using Item1, Item2 and so on.

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
```diff
@@ -83,52 +84,51 @@ public EventHubOutput(IConfiguration configuration, IHealthReporter healthReport
...
-                    // start a new batch.
-                    if (batches.Count == 0 ||
-                        batchByteSize + messageSize > EventHubMessageSizeLimit)
+                    void SendBatch(List<Tuple<MessagingEventData, int>> batch)
                     {
-                        batches.Add(new List<MessagingEventData>());
-                        batchByteSize = 0;
+                        // Since event hub limits each message/batch to be a certain size, we need to
+                        // keep checking the size in bytes of the batch and recursively keep splitting into two batches as needed
+                        if (batch.Count >= 2 && batch.Sum(b => b.Item2) > EventHubMessageSizeLimit)
```
karolz-ms: Noo :-) The `batch.Count >= 2` is a wrong thing to do--you DO WANT the batches to be as big as `EventHubMessageSizeLimit` allows

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
```diff
@@ -83,52 +84,51 @@ public EventHubOutput(IConfiguration configuration, IHealthReporter healthReport
...
+                        {
+                            var indexMiddle = batch.Count / 2;
+                            SendBatch(batch.Take(indexMiddle).ToList());
```
karolz-ms: Now with the extensive use of Linq you are re-calculating sizes of candidate batches and iterating over the same elements many times. I think it is better to just keep taking elements from the partition until you exceed the batch size limit, send the batch, start a new one and keep going until you run out of partition elements. It is both easier to understand and you consider each partition element just once.

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/EventHubOutputTests.cs
```diff
@@ -55,26 +55,91 @@ public void UsesIsoDateFormat()
...
+
+            Func<IEnumerable<MessagingEventData>, bool> verifyBatch = batch =>
+            {
+                if (batch.Count() != 1) return false;
+                return true;
+            };
+
+            client.Verify(c => c.SendAsync(It.Is<IEnumerable<MessagingEventData>>(b => verifyBatch(b))), Times.Once);
+            client.Verify(c => c.SendAsync(It.Is<IEnumerable<MessagingEventData>>(b => verifyBatch(b)), "partition1"), Times.Once);
+            client.Verify(c => c.SendAsync(It.Is<IEnumerable<MessagingEventData>>(b => verifyBatch(b)), "partition2"), Times.Exactly(3));
```
karolz-ms: I would not expect this. I would expect to have one call to SendEventsAsync(batch, "partition2") with 3 elements in the batch.

tedvanderveen: @karolz-ms I setup the three payloads in those events to contain 150k+ bytes of data, so the EventHubMessageSizeLimit kicks in and makes sure these events are outputted to Event Hub in single item batches :)

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
tedvanderveen: @karolz-ms I implemented named tuple members, thanks for this suggestion

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/EventHubOutputTests.cs
karolz-ms: @tedvanderveen ah, OK, that makes more sense. Still, I think it is important to verify that what should be batched is actually batched, because that is a key performance optimization that lets us send thousands of events/sec to EventHub from a single service process. So either this test, or a separate test should verify this.

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
tedvanderveen: @karolz-ms I added some additional comments in the code. What I if a batch is too big in total byte size, is to split the batch into two new batches each containing half of the original batch items. And keep doing this until the split-up batch is not too big anymore (and still contains at least 2 items, so it's even possible to split the batch up, hence the `batch.Count >= 2` evaluation.
If you do not appreciate this approach, let me know and I shall do the `foreach` approach.

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/EventHubOutputTests.cs
tedvanderveen: @karolz-ms I shall add an additional test for this!

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
karolz-ms: Sorry, I misread the code. It is fine, never mind :-*)

tedvanderveen:
@karolz-ms Please let me know if you require any additional changes.

PR #225, opened 2018-05-24
Test improvements

karolz-ms:
The tests have gotten really slow in the last couple of months. This PR speeds them up while hopefully retaining the current reliability (which was good, event on slow machines)


karolz-ms:
Thanks, Saar :)

PR #228, opened 2018-05-28
Added ability so select different connection types.

jeremysmith1:
This change will allow users to select a connection pool type with configuration. 

I removed the "useSniffingConnectionPooling" as I thought it would be confusing to have that setting with the connection type. Suggestions on how to rework this piece is desired. 

I also moved the logic for creating the connection into extensions methods and tested. Based of comments from @xiaomi7732 .


README.md
```diff
@@ -614,13 +615,14 @@ This output writes data to the [Elasticsearch](https://www.elastic.co/products/e
 | `type` | "ElasticSearch" | Yes | Specifies the output type. For this output, it must be "ElasticSearch". |
 | `indexNamePrefix` | string | No | Specifies the prefix to be used when creating the Elasticsearch index. This prefix, together with the date of when the data was generated, will be used to form the name of the Elasticsearch index. If not specified, a prefix will not be used. |
 | `serviceUri` | URL:port | Yes | Specifies where the Elasticsearch cluster is. This is needed for EventFlow to locate the cluster and send the data. Single URL or semicolon-separated URLs for connection pooling are accepted |
+| `connectionPoolType` | "SingleNode", "Static", "Sniffing", or "Sticky" | Yes | Specifies the Connection Pool that takes care of registering what nodes there are in the cluster. |
```
karolz-ms: 1. How about we do not support SingleNode at all? I think if you just use one URI with a Static connection pool, that is effectively the same as a SingleNode pool
2. Let's default to Static connection pool if it is not specified otherwise in the configuration

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Elasticsearch.Net;
+using Microsoft.Diagnostics.EventFlow.Configuration;
+
+namespace Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch
+{
+    internal static class ElasticSearchExtensions
```
karolz-ms: Is there a particular reason why you used extension methods instead of just adding them to `ElasticSearchOutputConfiguration` class?

If the answer is "no", I suggest to keep things simple

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+            this ElasticSearchOutputConfiguration elasticConfig, IHealthReporter healthReporter)
+        {
+            var enumParseResult =
+                Enum.TryParse<ElasticConnectionPoolType>(elasticConfig.ConnectionPoolType, out var elasticConnectionPool);
+            if (enumParseResult)
+            {
+                return elasticConnectionPool;
+            }
+
+            //Invalid config string report and throw
```
karolz-ms: Comment inaccurate--the code is not throwing here

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+            var errorMessage = $"{nameof(ElasticSearchOutput)}:  required 'ConnectionPoolType' configuration parameter is invalid. Using default.";
```
karolz-ms: "using default (static connection pool)" 

... and make it a warning, not a problem

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+            healthReporter.ReportProblem(errorMessage, EventFlowContextIdentifiers.Configuration);
+            return ElasticConnectionPoolType.SingleNode;
```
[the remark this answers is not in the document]
karolz-ms: I respectfully disagree. Questions to consider:
- How many ES clusters used in production are single node?
- Do we really want to have a default that causes EventFlow to stop sending logs when one node of the cluster goes down?

README.md
jeremysmith1: Looking through the elastic documentation does not really give you an answer. Looking at the code does though. 

A SingleNode connection does not allow you to use the ElasticClient.Ping(). 

Since the ElasticClient is not exposed to users, not being able to ping does not seem like an issue. So in short, I agree. I will remove the singleNode option.

On point 2, if we make "Static" the default, should we make this a optional configuration value? I would think yes, to remove the need for this elastic domain  knowledge.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchExtensions.cs
jeremysmith1: Sounds good to me. My thinking was that the configuration class was meant to be a POCO of some type. I will move these into the suggested class.

README.md
karolz-ms: Yes, let's make it optional

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchExtensions.cs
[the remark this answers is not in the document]
jeremysmith1: I agree, good point. It would be hard to tell as a user that you badly configured your multi node cluster. 

I will be making the Static connection pool the default. Also making the connection pool optional. And reflect this in the documentation.

jeremysmith1:
Awesome. Glad I could help. =)

Issue #219, opened 2018-05-23
ETW and EventSource inputs ignore EventCounter


Expecho:
I created an EventSource with some EventCounters as outlined in [this example](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Tracing/documentation/EventCounterTutorial.md) but upon testing I noticed EventFlow not picking up the values. Is this a known issue?

I checked the output using PerfView and it did collect the values of the 2 EventCounters.

My code:

	void Main()
	{
		using (var pipeline = DiagnosticPipelineFactory.CreatePipeline(@"d:\temp\Config.json"))
		{
	
			var input = Console.ReadLine();
			while (input != "q")
			{
				ServiceEventSource.Log.Request(float.Parse(input));
				ServiceEventSource.Log.Request2(float.Parse(input));
				input = Console.ReadLine();
			}
		}
	}
	
	[EventSource(Name = "My-EventSource")]
	public class ServiceEventSource : EventSource
	{
		public static ServiceEventSource Log = new ServiceEventSource();
		private EventCounter counter;
		private EventCounter counter2;
	
		private ServiceEventSource()
		{
			counter = new EventCounter("RequestCounter", this);
			counter2 = new EventCounter("RequestCounter2", this);
		}
	
		[NonEvent]
		public void Request(float count)
		{
			this.counter.WriteMetric(count);
		}
	
		[Event(1)]
		public void Request2(float count)
		{
			WriteEvent(1, count);
			this.counter2.WriteMetric(count);
		}
	}

My config:

     {
        "inputs": [
            {
                "type": "EventSource",
                "sources": [
                    {
                        "providerName": "My-EventSource"
                    }
                ]
            }
        ],
        "outputs": [
            {
                "type": "StdOutput",
            }
        ],
        "healthReporter": {
            "type": "CsvHealthReporter",
            "logFileFolder": "d:\\temp\\",
            "logFilePrefix": "HealthReport",
            "minReportLevel": "Warning",
            "throttlingPeriodMsec": "1000",
            "singleLogFileMaximumSizeInMBytes": "8192",
            "logRetentionInDays": "30",
            "ensureOutputCanBeSaved": "false"
        },
        "schemaVersion": "2016-08-11"
    }

karolz-ms:
Hi @Expecho 

I have run your program and I can see the regular event from the EventSource (event 1, associated with `Request2` method) is being captured properly

[...]

PR #229, opened 2018-05-30
Support EventCounters in EventSource Input

Expecho:
Implementation of support for EventCounters in EventSource as outlined in #219

### DIFF
--- a/test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/EventSourceInputTests.cs
+++ b/test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/EventSourceInputTests.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Tracing;
 using System.Linq;
+using System.Threading.Tasks;
 using Moq;
 using Xunit;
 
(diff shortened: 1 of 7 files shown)


test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/EventSourceInputTests.cs
```diff
@@ -183,6 +185,100 @@ public void CapturesEventsFromSourcesIdentifiedByNamePrefix()
...
+            });
+
+            using (eventSourceInput.Subscribe(observer.Object))
+            {
+                EventSourceInputTestSource.Log.ReportTestMetric(5);
+                EventSourceInputTestSource.Log.ReportTestMetric(1);
+
+                var firstTaskToComplete = await Task.WhenAny(
+                    testTaskCompletionSource.Task,
+                    Task.Delay(TimeSpan.FromSeconds(1.5))
```
karolz-ms: Some of the machines used for testing can be highly loaded and/or very slow, so please make the timeout larger, say, 3 seconds.

test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/EventSourceInputTests.cs
```diff
@@ -183,6 +185,100 @@ public void CapturesEventsFromSourcesIdentifiedByNamePrefix()
...
+                );
+
+                Assert.Equal(testTaskCompletionSource.Task, firstTaskToComplete);
+
+                var data = await testTaskCompletionSource.Task;
```
karolz-ms: ... or `var data = testTaskCompletionSource.Task.Result;`

test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/EventSourceInputTests.cs
karolz-ms: (applies to other tests as well)

PR #239, opened 2018-07-13
Add process info to Etw event payload

frblondin:
The Etw [TraceEvent](https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/TraceEvent.cs) contains [ProcessID](https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/TraceEvent.cs#L739) and [ProcessName](https://github.com/Microsoft/perfview/blob/dcece4936e6c8c60073eddd2213b00bca83aab69/src/TraceEvent/TraceEvent.cs#L754) properties. Having access to the process data seems important to me 😄.

This PR simply adds these two property values to the payload.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/TraceEventExtensions.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/TraceEventExtensions.cs
@@ -38,6 +38,8 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
             {
                 payloadData.Add(nameof(traceEvent.RelatedActivityID), traceEvent.RelatedActivityID.ToString());
             }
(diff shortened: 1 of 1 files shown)


karolz-ms:
Hey @frblondin thank you for your contribution!

Before we merge this change, I would like to understand a bit better why these two properties are useful to you. For example, are you analyzing ETW trace files and trying to correlate events you obtain via EventFlow with what you see in a trace file?

frblondin:
Thanks for your quick answer. I'm adding traces to measure advanced GC metrics that are only available as ETW events. I'm only interested in getting this data for specific processes.

I was missing the process info as they aren't necessarily provided by the event trace I'm subscribing to.

karolz-ms:
ETW input package version 1.3.3, with this change, should be available from Nuget shortly. Thanks again for the change!

frblondin:
Thanks!

PR #258, opened 2018-08-18
NLog Target as EventFlow Input

snakefoot:
Implemented like the Serilog-Sink. So manually registration of the created NLog-target in the NLog-Configuration is required.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.NLog/NLogInput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.NLog/NLogInput.cs
@@ -1,0 +1,263 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Diagnostics.EventFlow.Metadata;
+using NLog;
+using NLog.Config;
+using NLog.Layouts;
+using NLog.Targets;
+
+namespace Microsoft.Diagnostics.EventFlow.Inputs
+{
+    [Target(NLogTargetTypeName)]
+    public class NLogInput : TargetWithContext, IObservable<EventData>
+    {
+        internal const string NLogTargetTypeName = "EventFlowInput";
+
+        private static readonly IDictionary<NLog.LogLevel, LogLevel> ToLogLevel = new Dictionary<NLog.LogLevel, LogLevel>
+        {
+            [NLog.LogLevel.Trace] = LogLevel.Verbose,
+            [NLog.LogLevel.Debug] = LogLevel.Verbose,
+            [NLog.LogLevel.Info] = LogLevel.Informational,
+            [NLog.LogLevel.Warn] = LogLevel.Warning,
+            [NLog.LogLevel.Error] = LogLevel.Error,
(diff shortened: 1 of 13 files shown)


src/Microsoft.Diagnostics.EventFlow.Inputs.NLog/NLogInput.cs
```diff
@@ -0,0 +1,265 @@
...
+            Layout = "${message}";
+
+            RegisterNLogTargetType(healthReporter, GetType());
+
+            _eventFlowSubject = new EventFlowSubject<EventData>();
+        }
+
+        protected override void CloseTarget()
+        {
+            _eventFlowSubject?.Dispose();
```
karolz-ms: Just curious: do we ever expect the `_eventFlowSubject` to be null?

If not, then ordinary "dot" operator for function call is sufficient

src/Microsoft.Diagnostics.EventFlow.Inputs.NLog/NLogInput.cs
```diff
@@ -0,0 +1,265 @@
...
+                eventData.AddPayloadProperty("Message", RenderLogEvent(Layout, logEvent), _healthReporter, nameof(NLogInput));
+            }
+            catch (Exception ex)
+            {
+                NLog.Common.InternalLogger.Warn(ex, NLogTargetTypeName + "(Name={0}): Failed to render event message", Name);
+                _healthReporter.ReportWarning($"{nameof(NLogInput)}: event message could not be rendered{Environment.NewLine}{ex.ToString()}");
+                eventData.AddPayloadProperty("Message", logEvent.FormattedMessage, _healthReporter, nameof(NLogInput));
+            }
+
+            // MessageTemplate is always present on Serilog events
```
karolz-ms: .. but this is NLog input. Is it also the NLog convention to always include the message template?

If so, at leas the comment should be updated.

src/Microsoft.Diagnostics.EventFlow.Inputs.NLog/NLogInput.cs
```diff
@@ -0,0 +1,265 @@
...
+            {
+                foreach (var property in properties)
+                {
+                    try
+                    {
+                        eventData.AddPayloadProperty(property.Key, property.Value, _healthReporter, nameof(NLogInput));
+                    }
+                    catch (Exception ex)
+                    {
+                        NLog.Common.InternalLogger.Warn(ex, NLogTargetTypeName + "(Name={0}): Failed to render event property: {1}", Name, property.Key);
```
karolz-ms: This code that emits warnings about property rendering is very similar to code on lines 99-100, 161-162, 179-180 and 200-201. I suggest to create a helper method for this.

karolz-ms:
@snakefoot very much appreciate your contribution, thank you!!

There is a few things that need to be added before we merge:
1. Usage documentation for this input in README.md 
1. To easily instantiate the input, the `DiagnosticPipelineFactory` in the Core package needs to know about the factory for this input (which means an `NLogInputFactory` needs to be added as well).
1. [The assembly signing project](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Signing/Microsoft.Diagnostics.EventFlow.Signing.csproj) needs to be updated to account for the new NLog input assembly

The last item is somewhat gnarly in the sense that there is no way for you to test your change, so just let me know and I can take care of that for you.

karolz-ms:
@snakefoot the package has been published on Nuget. Please try it out and let me know if there are any problems.

Many thanks for the contribution!! :)

PR #260, opened 2018-08-20
Address issue 255

karolz-ms:
1. Create health reporter file in the current working directory by default
2. Handle IOExceptions from log file manipulation (expected to happen if the filesystem is read-only)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/HealthReporters/CsvHealthReporter.cs
```diff
@@ -231,19 +234,19 @@ public string RotateLogFile(string logFileFolder)
...
         {
             VerifyObjectIsNotDisposed();
             this.innerReportWriter?.Invoke(HealthReportLevel.Error, description, context);
         }
 
-        public void ReportWarning(string description, string context = null)
+        public virtual void ReportWarning(string description, string context = null)
         {
             VerifyObjectIsNotDisposed();
             this.innerReportWriter?.Invoke(HealthReportLevel.Warning, description, context);
```
xiaomi7732: Do I read it correct the intention is to support unit tests?
Per SHY, I will be cautious _virtual_ these methods - once open, there's no way close it. Is it possible to create a mock object directly from the interface for testing? Or maybe, overwrite SetNewStreamWrtier() directly.
Personally, I don't think it worth opening unless there's something critical I missed.
But on the other hand, I don't see anything wrong functionally.
I think it is your call :-)

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/HealthReporters/CsvHealthReporter.cs
```diff
@@ -507,10 +521,10 @@ private void ProcessLogFileFolder()
             }
             else
             {
-                basePath = Path.GetDirectoryName(System.Reflection.Assembly.GetAssembly(this.GetType()).Location);
+                basePath = Directory.GetCurrentDirectory();
```
xiaomi7732: I like this!

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/HealthReporters/CsvHealthReporter.cs
karolz-ms: It the motivation for this change was the new unit test--I wanted to verify that `ReportHealthy` implementation in the `CsvHealthReporter` works. Looking at it more closely I realize that I do not actually need the method to be virtual. I will revert these changes.

Issue #261, opened 2018-08-23
.net framework 4.7.2 uses .net framework 4.5.1 nuget package


ghelyar:
I have a .net framework 4.7.2 project that references both [Microsoft.Extensions.Configuration](https://www.nuget.org/packages/Microsoft.Extensions.Configuration/) 2.1.1 (a requirement of other packages) and other .net framework projects.

I want to use [Microsoft.Diagnostics.EventFlow.Core](https://www.nuget.org/packages/Microsoft.Diagnostics.EventFlow.Core/).

The .net framework 4.5.1 version of this package has a dependency restriction Microsoft.Extensions.Configuration.Abstractions (>= 1.1.1 && < 2.0.0)

The .net standard 2.0 version of this package has a dependency restriction Microsoft.Extensions.Configuration.Abstractions (>= 2.0.1)

.net framework 4.7.2 can use .net standard 2.0 packages, which would work in this case, but it picks the .net framework 4.5.1 package, so I can't use it.

As a simple reproduction, an asp.net core 2.1 project, targeting `net472`, just can't reference EventFlow at all.

Is it possible to force the project to reference the .net standard 2.0 version of the package? I would really not like to have to reference a dll directly.

If not, could a .net 4.7+ target be added to the package (and any related packages, like inputs and outputs, if they have the same problem) at some point that mirrors the .net standard 2.0 target's dependencies?

karolz-ms:
Would it be possible for you to just switch to .NET Core 2.1 framework? If not, why?

We could add a 4.7 target to EventFlow packages like you said... and settle on Ms.Ex.Configuration 2.x for that target. Would you consider submitting a PR?

ghelyar:
As there are references to other .net framework projects, some of which will likely never be net standard/core compatible, it can't just target .net core.

I will submit a PR as soon as I can.

karolz-ms:
The packages (Core 1.5.0, EtwUtilities 1.6.0 and ServiceFabric 1.5.0) should be available from nuget.org.

Thank you for helping with this change, much appreciated.

PR #263, opened 2018-08-24
Add net47 target framework

ghelyar:
Add a net47 target that uses the same package versions as the netstandard2.0 target. #261

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -1,5 +1,8 @@
 * Prerequisites
     * Visual Studio 2017
+	* .NET Framework 4.7.1 SDK and Targeting Pack
+	* .NET Core 1.0 SDK
+	* .NET Core 2.0 SDK
 
 * Build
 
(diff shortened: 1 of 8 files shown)


karolz-ms:
Thank you for the addition @ghelyar , much appreciated!

A few questions and suggestions:
1. Are you sure you do not want to have some of the inputs or outputs target net47 as well?
1. The [unit tests](https://github.com/Azure/diagnostics-eventflow/tree/master/test) should also have a run that targets net47... at least the Core assembly. It would also be nice to make a note in the [contribution guide](https://github.com/Azure/diagnostics-eventflow/blob/master/contributing.md) that .NET 4.7 SDK (targeting pack) is now required (and .NET Core 1.1 and 2.0 SDKs as well).
1. You will also need to modify the [assembly signing project](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Signing/Microsoft.Diagnostics.EventFlow.Signing.csproj) to account for new assemblies being built. So far we had 3 groups of assemblies there: "full", netstandard1.1 and netstandard2.0. This will no longer be sufficient, as we now have packages that have a binary for both net451 and net47 frameworks. I suggest creating a separate folder for assemblies targeting net47, e.g. named "full47". Let me know if you need help with this one.

ghelyar:
Thanks for the feedback. I wasn't sure if e.g. the test project should have the targets updated as well. I'll update this on Tuesday 28.

ghelyar:
After some further reading at https://github.com/dotnet/standard/issues/514 I'm thinking of bumping this up to net471. What do you think?

karolz-ms:
Yes, I think targeting 4.7.1 makes a lot of sense. Thanks for your diligence, appreciate it!

PR #265, opened 2018-09-10
Splunk output

hortha:
I've added a Splunk Output which utilises the Splunk HTTP Event Collector API.
Hopefully others will find it useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -838,11 +838,12 @@ ingestion service). |
 #### Splunk
 *Nuget Package*: [**Microsoft.Diagnostics.EventFlow.Outputs.Splunk**](https://www.nuget.org/packages/Microsoft.Diagnostics.EventFlow.Outputs.Splunk/)
 
-This output writes data to a Splunk HTTP Event Collector (HEC) (http://docs.splunk.com/Documentation/Splunk/latest/Data/AboutHEC). Here is an example showing all possible settings:
+This output writes data to a [Splunk HTTP Event Collector (HEC)](http://docs.splunk.com/Documentation/Splunk/latest/Data/AboutHEC).
+Here is an example showing all possible settings:
 ```json
 {
     "type": "Splunk",
-    "serviceBaseAddress": "https://mysplunkloadbalancer.com:8088/",
+    "serviceBaseAddress": "https://hec.mysplunkserver.com:8088/",
     "authenticationToken": "B5A79AAD-D822-46CC-80D1-819F80D7BFB0",
     "host": "localhost",
     "index": "main",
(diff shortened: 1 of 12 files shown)


yantang-msft:
@karolz-ms Just found couple other custom extensions:
[Google Big Query](https://github.com/tsu1980/diagnostics-eventflow-bigquery)
[SQL Server](https://github.com/ycamargo/diagnostics-eventflow-sqloutput)
[ReflectInsight](https://github.com/reflectsoftware/Microsoft.Diagnostics.EventFlow.Outputs.ReflectInsight)

karolz-ms:
@yantang-msft ha, thanks for setting me straight! I need to add a section to the docs then, esp. for the ReflectInsight folks who worked with us briefly last year.

hortha:
@karolz-ms & @yantang-msft thanks for the comments & information.

Ok I guess a new repo it is.. I notice that the ReflectInsight extensions are using a Microsoft namespace but I presume you wouldn't be too happy about that from a non Microsoft repo? 😀

I've got a few other bits I've been using around a different version of the ILogger input too - mostly so that context related data gets serialised to JSON properly (i.e. rather than embedded in the message) & indexed by Splunk (i.e. all the keys in a JSON document are indexed)

I'll get something sorted within the next week.

Thanks again for the feedback

karolz-ms:
@hortha sounds great.

Regarding the namespace part, you can argue both ways. Using non-Microsoft namespace eliminates the potential for naming conflicts, but on the other hand, sticking with Microsoft namespaces requires less namespaces to be imported for the user of the output, especially if the pipeline is created pro grammatically (not from configuration). I have a slight preference for non-Ms namespace but it is your pick.

Regarding JSON serialization, consider "manual" serialization using `JsonTextWriter` as opposed to `JsonConvert.SerializeObject`. The former not only gives you more control (arguably at the expense of extra code), but is at least an order of magnitude more performant, because it does not rely on reflection.

karolz-ms:
Closing this one--assume separate repo is the plan of record

hortha:
Hi @karolz-ms & @yantang-msft,

I'm finally getting around to creating that separate repository (in between the day job etc!).
I have a question for you about the MIT licensing (which I've struggled to find a definitive answer for on the internet).

The code I've written (not all of which was in this pull request) uses a mixture of re-namespaced classes (e.g. using classes from one of the other eventflow projects which were marked internal), tweaked versions of existing eventflow classes (e.g. an alternative Microsoft Logging Input implementation + my Splunk Output implementation based on a mixture of the existing eventflow output code) and brand new classes which I've written from scratch.

How would you (Microsoft) want this all attributed?

I was thinking of putting the following in the main LICENSE file:

```
MIT License

Copyright (c) 2018 Andrew Horth
Copyright (c) 2017 Microsoft Azure
```
...

and the following in the individual code files which were based on existing files:

```
// ------------------------------------------------------------------------------------------------
//  Copyright (c) Andrew Horth.  All rights reserved.
//  Copyright (c) Microsoft Corporation.  All rights reserved.
//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
// -------------------------------------------------------------------------------------------------
```

Would this be acceptable or can you suggest an alternative approach?

Thanks,

Andrew

karolz-ms:
Wonderful, thanks @hortha!

What you proposed for the attribution makes perfect sense and is exactly what our legal department 😬 recommends.

Looking at your implementation so far, is there something in EventFlow that we could tweak/open up to make developing 3rd party extensions like yours easier?

hortha:
In the Splunk Output I am making use of the internal HttpExponentialRetryMessageHandler class copied from the Microsoft.Diagnostics.EventFlow.Outputs.Oms project.
Rather than just making this public maybe it could be moved to the Core assembly and used by the HttpOutput as well as the OMS one? 
Also it has some TODO comments about implementing a circuit breaker and possibly it should be more configurable to allow for more than 3 attempts at a retry (i.e. a configurable amount)?

I will let you know of any other elements which I end up having to copy and change (i.e. rather than extending in some way).

hortha:
Hello again @karolz-ms,
I think I've finally got something pretty solid with a fair degree of code coverage.
I'm continuing to work on it but please let me know what you think so far.
The repository is at https://github.com/hortha/diagnostics-eventflow-splunk and I also uploaded a NuGet package.

karolz-ms:
Hi @hortha, overall looks very good! Just one question and one suggestion.

Question: just curious why you are using DI in https://github.com/hortha/diagnostics-eventflow-splunk/blob/master/src/Spotsoft.Diagnostics.EventFlow.Outputs.Splunk/SplunkOutputFactory.cs, instead of creating the client and other objects directly? Not that it is wrong or anything, but just not something that I would expect.

Suggestion: consider using [JsonWriter](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonWriter.htm#!) instead of `JsonConvert.SerializeObject` in https://github.com/hortha/diagnostics-eventflow-splunk/blob/master/src/Spotsoft.Diagnostics.EventFlow.Outputs.Splunk/Http/SplunkHttpEventCollectorClient.cs. I know that we use `JsonConvert.SerializeObject` in other places in EventFlow, but frankly, it is poor choice: it is an order of magnitude slower than `JsonWriter` because it is using reflection. 

Hope this helps! Thanks for this addition to EventFlow!
Karol

hortha:
Hi @karolz-ms thanks for taking the time to review my code.

I was using DI (ServiceCollection / ServiceProvider etc.) because my next enhancement is to target .NET Core 2.1 and use IHttpClientFactory along with Polly for exponential back off / circuit breaker (possibly) etc. and the code samples for that I was looking at were using ServiceCollection extension methods and using this sort of pattern (e.g. https://github.com/aspnet/HttpClientFactory/blob/master/samples/HttpClientFactorySample/Program.cs).
I just wanted to make the code targeting the other frameworks look as similar as possible and not necessarily have to use too many conditional compilation sections.
I'll review again once I've added that functionality as I understand what you are saying.

As for using `JsonWriter`, I wasn't sure how that would serialise the `Payload` property of the `EventData` class e.g. I want to use this to contain context information which could be anonymous classes, or collections / dictionaries themselves (or contain other nested classes etc.).
My requirement is to be able fully serialise the Payload to valid JSON so that each of the key / values & nested key / values are indexed in Splunk.
`JsonConvert.SerializeObject` seems to auto-magically do what I want, but I am willing to put some effort into seeing if `JsonWriter` can produce the same results (but faster!).

Thanks again for the feedback!

karolz-ms:
@hortha good point about the `JsonWriter`--I kind of forgot that the `Payload` can handle arbitrary data. My comment was too simplistic, sorry about that.

In general case `EventData` instances will be carrying arbitrary objects, and to properly serialize them you have to use reflection, which is exactly what `JsonConvert.SerializeObject` is doing.

But in many practical cases the data will be just simple types: numbers and strings, maybe Guids or boolean values. If this is true, calling `ToString` on a value might give you what you need, and combined with `JsonWriter` is much faster than `JsonConvert.SerializeObject`. So I would *consider* having a "fast mode" that does just that; enabled by a configuration option. This is just a suggestion, not a must-have by any means.

See https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs#L210 for an example of this approach in ASP.NET `EventSourceLogger` implementation.

hortha:
@karolz-ms thanks for that link - interesting to see that.

I think maybe there is a hybrid approach which may still give speed benefits - maybe I can use reflection just to serialise the arbitrary Payload data?
I'll give it a go over the next few days..

PR #268, opened 2018-09-15
Update EventFlow.Inputs.PerformanceCounter to support .NET standard 2.0 

tomheijmans:
Updated the package Microsoft.Diagnostics.EventFlow.Inputs.PerformanceCounter to support .NET standard 2.0. It uses the System.Diagnostics.PerformanceCounter nuget package which is released by Microsoft.

Minor package version update to 1.4.0


tomheijmans:
Thanks Karol for your feedback. Didn't know we could do that. Fixed it so it will support .NET 4.5.1 back again.

Issue #269, opened 2018-09-21
New payload properties generated for fields with same name regardless of value


reddogaw:
We have some log messages whom add payload fields that overlap names (and values) with payload properties added from an other scope. This leads to generated names being appended with an "_" and semi-random index number.

Could/should we be additionally checking for value equality here? If they have the same name and value as existing then just let it be.

```
using (logger.BeginScope("Username:{username}", "test")
{
    logger.LogInformation("Found user with username {username}", "test");
}
```

Currently that would result in payload properties with names:
* username
* username_1

Ideally, due to the equality of the key and value, it'd be good just have a single payload property added. If they had different values, definitely keep generating unique keys.

I think this is due to logic here: 
https://github.com/Azure/diagnostics-eventflow/blob/6fca5e00d31584445568bc98253db3966b56358a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/PayloadDictionaryUtilities.cs#L19-L23

karolz-ms:
Sounds like a good idea. Would you be willing to submit a PR?

PR #272, opened 2018-09-24
Avoid adding duplicate equivalent payload properties

reddogaw:
Fixes:  #269 

When duplicate payload properties are added, if the new value equals the existing value then it's not added to the payload at all and a health warning is written.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/PayloadDictionaryUtilities.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/PayloadDictionaryUtilities.cs
@@ -16,24 +16,40 @@ namespace Microsoft.Diagnostics.EventFlow
             Requires.NotNull(key, nameof(key));
             Requires.NotNull(healthReporter, nameof(healthReporter));
 
-            if (!payload.ContainsKey(key))
+            if (!payload.TryGetValue(key, out var existingValue))
             {
                 payload.Add(key, value);
+                return;
+            }
+            else if ((existingValue?.Equals(value)).GetValueOrDefault(false))
+            {
+                // Existing value with same key is equivalent to the input value
+                // We can return immediately to avoid adding duplicate key/value into the payload
+                healthReporter.ReportWarning($"The property with the key '{key}' already exist in the event payload with equivalent value. Value was not re-added", context);
                 return;
             }
 
             string newKey;
             int i = 1;
+
             //update property key till there is no such key in dict
             do
             {
                 newKey = key + "_" + i.ToString("d");
                 i++;
             }
-            while (payload.ContainsKey(newKey));
+            while (payload.TryGetValue(newKey, out existingValue) && !(existingValue?.Equals(value)).GetValueOrDefault(false));
 
-            payload.Add(newKey, value);
-            healthReporter.ReportWarning($"The property with the key '{key}' already exist in the event payload. Value was added under key '{newKey}'", context);
+            if (!payload.ContainsKey(newKey))
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/PayloadDictionaryUtilities.cs
```diff
@@ -16,24 +16,40 @@ public static void AddPayloadProperty(IDictionary<string, object> payload, strin
...
+            if (!payload.TryGetValue(key, out var existingValue))
             {
                 payload.Add(key, value);
                 return;
             }
+            else if ((existingValue?.Equals(value)).GetValueOrDefault(false))
+            {
+                // Existing value with same key is equivalent to the input value
+                // We can return immediately to avoid adding duplicate key/value into the payload
+                healthReporter.ReportWarning($"The property with the key '{key}' already exist in the event payload with equivalent value. Value was not re-added", context);
```
karolz-ms: Do not warn, just return. Emitting a warning has operational implications (the relevant component might be considered to be in "warning" state, requiring human intervention). In this case we might be trying to add some data more than once, but there is no data loss, so we still should be considered generally "healthy"

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/PayloadDictionaryUtilities.cs
```diff
@@ -16,24 +16,40 @@ public static void AddPayloadProperty(IDictionary<string, object> payload, strin
...
-            payload.Add(newKey, value);
-            healthReporter.ReportWarning($"The property with the key '{key}' already exist in the event payload. Value was added under key '{newKey}'", context);
+            if (!payload.ContainsKey(newKey))
+            {
+                payload.Add(newKey, value);
+                healthReporter.ReportWarning($"The property with the key '{key}' already exist in the event payload. Value was added under key '{newKey}'", context);
+            }
+            else
+            {
+                healthReporter.ReportWarning($"The property with the key '{key}' already exist in the event payload with equivalent value under key '{newKey}'. Value was not re-added", context);
```
karolz-ms: ... but keep this. If we get here (name clash at first attempt, but then we find an auto-generated property with the same value), something is fishy.

test/Microsoft.Diagnostics.EventFlow.Core.Tests/EventDataTests.cs
```diff
@@ -36,5 +37,92 @@ public void GetPropertyValueGetValueFromPayload()
...
+            var healthReporterMock = new Mock<IHealthReporter>();
+            var data = new EventData();
+            data.AddPayloadProperty("prop", "value", healthReporterMock.Object, "tests");
+            data.AddPayloadProperty("prop", "value", healthReporterMock.Object, "tests");
+
+            Assert.Equal(1, data.Payload.Keys.Count);
+            object value;
+            Assert.True(data.TryGetPropertyValue("prop", out value));
+            Assert.Equal("value", value);
+            healthReporterMock.Verify(m => m.ReportWarning(It.Is<string>(s => s.StartsWith("The property with the key 'prop' already exist in the event payload with equivalent value")), It.IsIn("tests")));
```
karolz-ms: As discussed above, I suggest we do not warn in this situation

reddogaw:
@karolz-ms - I've update with the changes requested. Thanks for the consideration.

karolz-ms:
@reddogaw Core package 1.5.1, containing this change, should be available from Nuget now :relaxed:

PR #281, opened 2018-10-25
Ensure that health reporter file is only created when there are any reports

karolz-ms:


karolz-ms:
Super, thank you @xiaomi7732 !

Issue #283, opened 2018-11-02
System.Diagnostics.DiagnosticSource input


ghelyar:
I was just wondering whether there was any existing [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource/) input, and whether it would be worth adding one? If so, I don't mind doing it when I have time.

`DiagnosticSource` seems to be quite similar to `EventSource`, but as it's in-process, it seems like a bit of a nicer fit for EventFlow.

karolz-ms:
@ghelyar your help would be much appreciated, as all members of the EventFlow team here at Microsoft are currently working primarily on other projects till at least mid-to-late January, myself included. 

It is true that `DiagnosticSource` is similar to `EventSource`, the main difference being that `EventSource` deals only with easily serializable types in the payload, vs. `DiagnosticSource` that can handle any data. Fortunately EventFlow is more like `DiagnosticSource` than `EventSource` in this respect, so stuffing `DiagnosticSource` event data into EventFlow `EventData` instance should be quite straightforward.

> BTW, both `EventSource` and `DiagnosticSource` are in-process, really, it is just that the default `EventListener` for `EventSource` is one that produces ETW traces on Windows, so many people think 
that producing ETW is the primary purpose of `EventSource`. It is not. `EventSource` is foremost a *logging API*, just like `DiagnosticSource`.

karolz-ms:
@ghelyar will you also be able to send a PR that updates the README with DiagnosticSource input information?

ghelyar:
Can you also please confirm that as the source has no level property, leaving the log level of these events as 0 is sufficient? This is described as [NotUsed](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/LogLevel.cs)

karolz-ms:
Well, there is no "error level" concept for DiagnosticSource events, so it makes sense to use `NotUsed`. It also makes sense that this is the highest level a.k.a. LogAlways in ETW, resulting in these events going through the filters.

karolz-ms:
I think the doc PR that you sent wraps it up. Again many thanks for this contribution @ghelyar !

PR #284, opened 2018-11-09
Add Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource

ghelyar:
Add an EventFlow input for System.Diagnostics.DiagnosticSource #283

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipelineFactory.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipelineFactory.cs
@@ -325,6 +325,7 @@ namespace Microsoft.Diagnostics.EventFlow
             inputFactories["ApplicationInsights"] = "Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsightsInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights";
             inputFactories["Log4net"] = "Microsoft.Diagnostics.EventFlow.Inputs.Log4netFactory, Microsoft.Diagnostics.EventFlow.Inputs.Log4net";
             inputFactories["NLog"] = "Microsoft.Diagnostics.EventFlow.Inputs.NLogInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.NLog";
+            inputFactories["DiagnosticSource"] = "Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource.DiagnosticSourceInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource";
 
             outputFactories = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             outputFactories["ApplicationInsights"] = "Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsightsOutputFactory, Microsoft.Diagnostics.EventFlow.Outputs.ApplicationInsights";
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Configuration/DiagnosticSourceConfiguration.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Configuration/DiagnosticSourceConfiguration.cs
@@ -1,6 +1,11 @@
-﻿namespace Microsoft.Diagnostics.EventFlow.Configuration
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.Diagnostics.EventFlow.Configuration
 {
-    public sealed class DiagnosticSourceConfiguration
+    public class DiagnosticSourceConfiguration
     {
         public string ProviderName { get; set; }
     }
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Configuration/DiagnosticSourceConfiguration.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Configuration/DiagnosticSourceConfiguration.cs
@@ -1,0 +1,7 @@
+﻿namespace Microsoft.Diagnostics.EventFlow.Configuration
+{
+    public sealed class DiagnosticSourceConfiguration
+    {
+        public string ProviderName { get; set; }
+    }
+}
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/DiagnosticListenerObserver.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/DiagnosticListenerObserver.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.Diagnostics.EventFlow.Configuration;
+
+namespace Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource
+{
+    internal sealed class DiagnosticListenerObserver : IObserver<DiagnosticListener>, IDisposable
+    {
+        private readonly IHealthReporter _healthReporter;
+        private readonly object _lock = new object();
+        private readonly IObserver<EventData> _output;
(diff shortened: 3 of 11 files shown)


ghelyar:
This is the implementation and tests, but:

- The csproj does not contain anything for nuget
- Nothing in particular has been done to get the code signed
- I'm not sure what needs to be done to be able to use the input without adding the assembly name to the `extensions` section of the json file.

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Configuration/DiagnosticSourceConfiguration.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.Diagnostics.EventFlow.Configuration
```
karolz-ms: In the main EventFlow repo all source files need to have Microsoft Copyright notice--see existing code

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/DiagnosticListenerObserver.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+        }
+
+        public void OnNext(DiagnosticListener listener)
+        {
+            if (_sources.Any(s => string.Equals(s.ProviderName, listener.Name, StringComparison.Ordinal)))
+            {
+                lock (_lock)
+                {
+                    _subscriptions.Add(listener.Subscribe(new EventObserver(listener.Name, _output, _healthReporter)));
```
karolz-ms: If the DiagnosticListenerObserver is disposed, should we still be adding subscriptions to the subscription list?

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/DiagnosticSourceInput.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    public sealed class DiagnosticSourceInput : IObservable<EventData>, IDisposable
+    {
+        private readonly DiagnosticListenerObserver _listenerObserver;
+        private readonly IDisposable _listenerSubscription;
+        private readonly EventFlowSubject<EventData> _subject;
+
+        public DiagnosticSourceInput(IReadOnlyCollection<DiagnosticSourceConfiguration> sources, IHealthReporter healthReporter)
+        {
+            if (sources == null)
```
karolz-ms: (minor) We use Validation package in other parts of EventFlow, so you can be a bit less verbose and say

```csharp
Requires.NotNull(sources, nameof(sources));
```
E.g. https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/EtwInput.cs#L31

(just a suggestion)

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Configuration/DiagnosticSourceConfiguration.cs
```diff
@@ -0,0 +1,7 @@
...
+{
+    public sealed class DiagnosticSourceConfiguration
```
karolz-ms: Please leave the classes unsealed if possible. I know some recommend to seal all classes by default, but we do not do that in EventFlow code.

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
karolz-ms: Please add the common properties we have on all parts of EventFlow--for an example see one of the input project like https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/Microsoft.Diagnostics.EventFlow.Inputs.Etw.csproj

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipelineFactory.cs
```diff
@@ -325,6 +325,7 @@ private static void ReportInvalidPipelineConfiguration(IHealthReporter healthRep
             inputFactories["ApplicationInsights"] = "Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsightsInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights";
             inputFactories["Log4net"] = "Microsoft.Diagnostics.EventFlow.Inputs.Log4netFactory, Microsoft.Diagnostics.EventFlow.Inputs.Log4net";
             inputFactories["NLog"] = "Microsoft.Diagnostics.EventFlow.Inputs.NLogInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.NLog";
+            inputFactories["DiagnosticSource"] = "Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource.DiagnosticSourceInputFactory, Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource";
```
karolz-ms: Please also increment the EventFlow.Core package version number in the EventFlow.Core project file

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFrameworks>netstandard1.6;netstandard2.0;net451;net471</TargetFrameworks>
```
karolz-ms: I do not see anything specific to net471 in your code. Do we need to target 4.7.1 specifically, or is net451 sufficient?

(we do have to target both netstandard1.6 and netstandard2.0 because of compatibility issues between .NET Core 1.1 and 2.x)

karolz-ms:
BTW, please do not worry about code signing and Nuget package creation, I will take care of these when the change is merged.

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/DiagnosticListenerObserver.cs
ghelyar: The `DiagnosticSourceInput` disposes of the subscription before disposing of the observer, so this shouldn't happen, but I've added a `disposed` flag for belt and braces.

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/DiagnosticSourceInput.cs
ghelyar: I'm not really a fan of such libraries. With the changes to the last few versions of the C# language, like `nameof` and being able to `throw` from `??` and `?:` they aren't as necessary, barely saving any code, and add extra dependencies and potentially supply chain attack vectors.

The gold partner ISV where I work is quite strict on which packages can be used in our software, and the dependency on the Validation package is the kind of reason they may reject Microsoft.Diagnostics.EventFlow.

Although there is already a dependency on Validation, I didn't want to make it any worse.

karolz-ms: OK, it is not a big deal

For the sake of your partner ISV: I also do not like unnecessary dependencies, and as one can see, EventFlow has not many. The Validation library has a minimalist design that I very much like. In addition to cutting down the amount of code, adopting Validation library improves readability of the code--it makes argument checks and assumptions very explicit. I also happen to know the author; he is a fellow Microsoft employee and a very conscientious developer. I have been using Validation for various project for many years and my concern about this library having negative impact on EventFlow is zero.

src/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource/Microsoft.Diagnostics.EventFlow.Inputs.DiagnosticSource.csproj
karolz-ms: OK, will review, thx

Issue #291, opened 2018-11-27
ApplicationInsights input does not set metadata correctly for processing by outputs


voltcode:
I am trying to develop a simple outputter that parses exception information.
I tested it with NLog input and calling logger.Error(exception);     works fine - exception is recognized as Exception in my output class.

however, when I use the application insights input it does not work correctly.

in the client code I call:  telemetryClient.TrackException(exception);   

in my output this condition is not fulfilled:
if (eventData.TryGetMetadata(ExceptionData.ExceptionMetadataKind, out metadataSet))
{
}

When inspecting the eventData in profiler, I can see that the private property - metadata is null when the input is from Application Insights, whereas from NLog it is not null.
In both NLog and AI cases, the payload is not null.

It seems like the data is flowing through but is generated correctly on the ai side.

I am using:
<package id="Microsoft.ApplicationInsights" version="2.8.1" targetFramework="net462" />
  <package id="Microsoft.Diagnostics.EventFlow.Core" version="1.4.1" targetFramework="net462" />
  <package id="Microsoft.Diagnostics.EventFlow.Inputs.ApplicationInsights" version="1.4.1" targetFramework="net462" />

karolz-ms:
Well, it is just a bug (not-implemented issue) in Application Insights input. 😐  It should be setting Dependency/Request/Metric/Exception metadata on events depending on input telemetry type, but it does none of these things today.

How important is this feature for you, @voltcode ?

voltcode:
Without being able to rely on AI input, I probably have to give up using Ai altogether, which is a shame because it means I may need to reimplement what it does by stiching other inputs together + replicating some of the Ai data generation myself.  I'd say that's pretty serious reason whether or not to use AI :) 

It may also be that because of the need for custom development, I may ditch using eventflow too, which would be a shame because otherwise it does a good job.

karolz-ms:
Adding metadata to events produced by Application Insights input is not a complicated fix. Would it be sufficient for your purposes to have it ready by the end of this week?

Also, if you could describe a bit what your system architecture is and where the logs are going, we might give you some tips on how to get most benefits from App Insights

voltcode:
That would be great, thanks!

Regarding architecture - requirements aren't finalized yet, but I need to be ready for some on-premise deployments without internet access for at least some of the services and be able to merge this data with existing logging infrastructure NLog, etc. as eventflow permits. 
[...]

PR #292, opened 2018-11-30
Make AppInsights input produce standard metadata

karolz-ms:


karolz-ms:
Addresses [issue 291](https://github.com/Azure/diagnostics-eventflow/issues/291)

PR #293, opened 2018-12-03
Log4Net input should not emit traces to console

karolz-ms:
Just noticed that Log4Net input will allow traces to be sent to Console.Out. This should not happen.


karolz-ms:
@jeremysmith1 please let me know what you think about this change

karolz-ms:
Hey, no worries, Jeremy! It is not a big deal. Thanks for a quick review!

PR #295, opened 2018-12-06
Move high precision timestamps from EventSource input to Core

ghelyar:
- Move high precision timestamps from EventSource input to Core
- Use high precision timestamps from DiagnosticSource input
- Give DiagnosticSource baggage items precedence over tags in the event of a conflict, to improve correlation across activities

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DateTimePrecise.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DateTimePrecise.cs
@@ -1,0 +1,39 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+#if NET451 || NET471
+using System.Runtime.InteropServices;
+#endif
+
+namespace Microsoft.Diagnostics.EventFlow
+{
+    public static class DateTimePrecise
+    {
+#if NET451 || NET471
+        [DllImport("Kernel32.dll", CallingConvention = CallingConvention.Winapi)]
+        private static extern void GetSystemTimePreciseAsFileTime(out long filetime);
+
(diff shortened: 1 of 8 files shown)


ghelyar:
This applies to all versions of .net framework, as the net46 enhancement only increased resolution from ~15ms to 1ms, while .net core 2.0 and `GetSystemTimePreciseAsFileTime` both give 0.0001ms resolution.

The existing test for EventSource only tested that over the course of 16ms, 8 different timestamps were produced, which doesn't test this level of precision.

I have added a new test which proves the 0.0001ms resolution when compiled into a console application targeting the different frameworks, but when run under the test runner with either xunit or nunit it seems to lose this resolution, so I've left the new test on 1.1ms with a comment.

karolz-ms:
@ghelyar thank you for this valuable contribution, much appreciated!

Good job trying to verify the expected time resolution via tests. It is notoriously tricky. The test runner is running tests in different test classes concurrently that affects the timing. It can be changed via attributes... but even that is often not sufficient, because of some other workloads on the test machine that may affect our test run. We'll see if the 1000-samples averaging is sufficiently robust; if not, we might need to try to eliminate the extremes from the sample set, or try some other tricks.

ghelyar:
@karolz-ms it wasn't just concurrency, it seems that as soon as the test runner or test framework gets involved, everything but netcore2.0 drop immediately from getting almost exactly 1 nanosecond resolution to getting almost exactly 1 millisecond, even if it's the only test in the suite. I tried nunit and xunit in visual studio test explorer, dotnet test, vstest.console and ncrunch all with the same results, but if I just make a console application and run the test code in a try/catch, it works perfectly, even if the console application and the test fixture are in the same assembly. Eventually I just left the test on < 1ms, with 10% tolerance so 1.1ms, over 1000 runs, which passes reliably but really it's not testing the resolution it's expecting.

I considered adding a console application and running it as a child process but it seemed like a sledgehammer to crack a nut.

karolz-ms:
@ghelyar makes sense

Nugets have been published

Issue #294, opened 2018-12-05
logging entries missing via OMS output


lnhzd:
We configured our pipeline to have dual outputs: OMS output and Std output.
We are expecting entries in these two outputs match up, however they do not correlate with each other: For a given period of time (a few hours) we are seeing 9 entries of error level logging in Stdoutput, while there is only one appearing in OMS. We noticed missing entries because we were after certain errors in OMS but they were not there, potentially the same issue may happen on other logging level, e.g. info, debug, as well.
the traffic was pretty low, we were not running any load tests whatever, a few smoke tests only.

{
  "inputs": [
    {
      "type": "EventSource",
      "sources": [
        {
          "providerName": "MyProviderName",
          "level": "Verbose"
        }
      ]
    }
  ],
  "outputs": [
    {
      "type": "StdOutput"
    },
    {
      "type": "OmsOutput",
      "workspaceId": "servicefabric:/ef/workspaceId",
      "workspaceKey": "servicefabric:/ef/workspaceKey",
      "logTypeName": "myTableName"
    }
  ],
  "schemaVersion": "2016-08-11"
}

lnhzd:
no it's only an intermittent error (may be more frequent than intermittent but definitely not happening on every single entry), this effectively means the warning of health report will be flushed by the next successful call to OMS presumably? we  didn't set up any alert, nor did anyone sitting there staring at sf explorer for a 24 hours so cannot really tell, unless there's a way to check historical data of health report?

it has been happening for a while, for at least two of our services so may not relate to the shortage you mentioned.thanks.

lnhzd:
Thanks Karolz, following your suggestion I passed in a custom health reporter that outputs to std output, and found that for a given period (20 min), out of the 100 ish requests go out to oms, 6 failed because "OmsOutput: OMS REST API returned an error. Code: Forbidden Description: Forbidden {"Error":"InvalidAuthorization","Message":"An invalid signature was specified in the Authorization header"}."

These failing requests are almost data-specific that contains a payload of json string with an already serialized object that contains exception stack trace, ex.ToString. 
Looked at the signatureInput, message.Length does look suspicious I wonder if it's something to do with how you validate message length when a message is received on your side?

I've got a raw request (goes to OMS) that can be used to demo the problem, but I'm not comfortable to paste the payload here, is there an email I can send the payload across for you to investigate the problem please? thanks.

lnhzd:
[...]

PR #296, opened 2018-12-07
Update OmsOutput.cs

lnhzd:
fix data-specific authorization failure: content-length must be in byte size

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/OmsOutput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/OmsOutput.cs
@@ -115,10 +115,9 @@ namespace Microsoft.Diagnostics.EventFlow.Outputs
                 string jsonData = JsonConvert.SerializeObject(events);
 
                 string dateString = DateTime.UtcNow.ToString("r");
-
-                string signature = BuildSignature(jsonData, dateString);
-
-                HttpContent content = new StringContent(jsonData, Encoding.UTF8, JsonContentId);
+                HttpContent content = new StringContent(jsonData, Encoding.UTF8);
+                string signature = this.BuildSignature(content.Headers.ContentLength, dateString);
+				
                 content.Headers.ContentType = new MediaTypeHeaderValue(JsonContentId);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/OmsOutput.cs
```diff
@@ -157,7 +157,8 @@ public async Task SendEventsAsync(IReadOnlyCollection<EventData> events, long tr
         private string BuildSignature(string message, string dateString)
         {
             string dateHeader = $"{MsDateHeaderName}:{dateString}";
-            string signatureInput = $"POST\n{message.Length}\n{JsonContentId}\n{dateHeader}\n{OmsDataUploadResource}";
+            byte[] messageInBytes = Encoding.UTF8.GetBytes(message);
```
karolz-ms: Minor perf improvement suggestion: swap lines 119 and 121 above and use `StringContent.TryComputeLength()` (will always succeed for StringContent) to compute the message length, then pass it as a (new) parameter to `BuildSignature`. This way we UTF8-encode the message just once instead of twice.

karolz-ms:
Fixes https://github.com/Azure/diagnostics-eventflow/issues/294

src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/OmsOutput.cs
lnhzd: @karolz-ms thanks, unless I missed anything TryComputerLength() is an internal method I'm unable to use? I've therefore used content.Headers.ContentLength instead presumably it calls TryComputerLength internally? Feel free to make any further changes/suggestions if required. thanks.

karolz-ms: @lnhzd sorry, my mistake, did not look at the signature carefully enough--but you found an alternative way to implement the suggestion, great job!

karolz-ms:
Nuget package 1.4.1 has been published

lnhzd:
thanks a lot

PR #301, opened 2019-01-18
Add ingest node default pipeline support to Elasticsearch output

yobryon:
Adds configuration parameter to Elasticsearch output config type, allowing user to specify a default pipeline to be used when creating a new index.  This corresponds to the [`index.default_pipeline` dynamic index setting](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/index-modules.html#dynamic-index-settings).

Having this specified during index creation ensures that, if pipeline usage is desired in a given use case, that all documents indexed are done via the pipeline from the inception of that index.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
@@ -388,6 +388,7 @@ namespace Microsoft.Diagnostics.EventFlow.Outputs
             indexSettings.Settings.NumberOfReplicas = this.connectionData.Configuration.NumberOfReplicas;
             indexSettings.Settings.NumberOfShards = this.connectionData.Configuration.NumberOfShards;
             indexSettings.Settings.Add("refresh_interval", this.connectionData.Configuration.RefreshInterval);
+            indexSettings.Settings.Add("default_pipeline", this.connectionData.Configuration.DefaultPipeline);
 
             ICreateIndexResponse createIndexResult = await esClient.CreateIndexAsync(indexName, c => c.InitializeUsing(indexSettings)).ConfigureAwait(false);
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -388,6 +388,7 @@ private async Task EnsureIndexExists(string indexName, ElasticClient esClient)
             indexSettings.Settings.NumberOfReplicas = this.connectionData.Configuration.NumberOfReplicas;
             indexSettings.Settings.NumberOfShards = this.connectionData.Configuration.NumberOfShards;
             indexSettings.Settings.Add("refresh_interval", this.connectionData.Configuration.RefreshInterval);
+            indexSettings.Settings.Add("default_pipeline", this.connectionData.Configuration.DefaultPipeline);
```
karolz-ms: Should we only set it if `Configuration.DefaultPipeline` is not null or empty?

yobryon: I thought about that, but I did a quick test of the Nest IndexSettings class and serializer, and found that if the dictionary entry had a null value, the field did not get rendered in the output, which I then confirmed with another quick test against a test elastic instance. 

It wouldn't hurt anything to wrap it in an if statement, I was just going for minimal additional code.

And thank you guys for this framework! It has been very helpful.

karolz-ms: Thank you!

Yes, makes sense, and I am glad that Nest is doing the right thing here, but let's be a bit more defensive in case they "change their mind" down the road ☺️

yobryon: Added another commit with a null check, such that this parameter will only be added to the index settings if the corresponding Configuration property has a non-null value.

karolz-ms:
Elasticsearch output 2.6.1 has been published on nuget.org

PR #315, opened 2019-04-01
Protected members

AssafTzurEl:
Make OmsOutput  fields healthReporter and connectionData protected, in order to allow access by OmsOutput subclasses.


src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/OmsOutput.cs
```diff
@@ -3,6 +3,11 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+using Microsoft.Diagnostics.EventFlow.Configuration;
```
karolz-ms: It does not really matter that much but we have a slight preference for keeping System namespaces above others. There is a setting in Visual Studio that helps maintain that order.

karolz-ms:
@AssafTzurEl please let me know if you need a new version of OmsOutput (with this change) to be published on Nuget

src/Microsoft.Diagnostics.EventFlow.Outputs.Oms/OmsOutput.cs
AssafTzurEl: @karolz-ms  Thanks. I'm using the "remove and sort" command in VS, and I like it - the sorting helps me locate using's faster.

AssafTzurEl:
@karolz-ms re/ new Nuget version: No rush.

Issue #311, opened 2019-03-18
Eventflow stops periodical output


NicolasPrivatViseo:
Hi,
We have a Service Fabric application which uses eventflow to send logs into an event hub and appInsights (and a HttpOuput for debug purpose).
At startup event are correctly sent in the outputs but after a while (not clearly defined) it stops outputting the events.

We dug into the code and saw that you are using TPL Dataflow with BatchBlock. We also noticed the timer mecanism to trigger the batch in the DiagnosticPipeline initialization. 
We noticed that if we reach the batch size (100 events), then eventflow sends the output.
But the timer (`batchTimer` in `DiagnosticPipeline.cs` class) should allow to send the events even if the batch is not completed as it works at startup. 
The timer is configured with the default 500ms delay. The timer still executes the function `TriggerBatch` but the events are not sent to outputs.

As we can see in the capture below, we have no postponed message, and 10 events in the input queue, that are not sent in the output queue of the `BatchBlock`

![image](https://user-images.githubusercontent.com/44289644/54547961-1f990980-49a7-11e9-9cfc-c2164a091c9e.png)

Have you encountered similar issue, and do you have advice on where to look at?

We are using .Net Core 2.1, we have cloned the eventflow core (master) repo locally to debug within our project.

karolz-ms:
Hi @NicolasPrivatViseo 

thanks for reporting this issue. This is first time we hear about it.

From what you are saying I assume that the EventFlow timer keeps calling `TriggerBatch` but, for some reason, the batch is not being actually triggered?

If so, it would be interesting to look at `BatchBlock` more closely. In particular, when you get to the state described above, could you check whether `_dataflowBlockOptions.CancellationToken.IsCancellationRequested` is true and what [`_target.BatchesNeedProcessing`](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Tasks.Dataflow/src/Blocks/BatchBlock.cs#L570) property returns?

NicolasPrivatViseo:
Hi @karolz-ms,
Thanks for the quick answer. Indeed the function `TriggerBatch` keeps being triggered.

It calls the [ProcessAsyncIfNecessary](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Tasks.Dataflow/src/Blocks/BatchBlock.cs#L623) method. The `BatchesNeedProcessing` property is false.

About `_dataflowBlockOptions.CancellationToken.IsCancellationRequested`, it is set to false 

[...]

PR #325, opened 2019-05-21
Improve batch timer handling in DiagnosticPipeline

karolz-ms:
This is an attempt to reduce the possibility of ending in a state described in issue 311 https://github.com/Azure/diagnostics-eventflow/issues/311 It is not entirely clear to me what the problem is there because I have a very hard time reproducing the issue, but the one time when it did repro with debugger attached, I saw that the batcher timer was not being fired. As far as I could tell there was nothing wrong with EventFlow pipeline.

I suspect this is an initialization timing issue that results in thread pool thred exhaustion and the timer proc not being scheduled as it should. One can get more or less to the same state when threadpool threads are manually suspended (the timer is no longer fired, and no new threads are added to the thread pool).

This change ensures that only one timer proc is ever running, so it should help, and in general is a good rule to follow.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Diagnostics.EventFlow
         private List<Task> outputCompletionTasks;
         private bool disposed;
         private bool disposeDependencies;
-        private object disposalLock;
+        private object batcherTimerDisposalLock;
         private CancellationTokenSource cancellationTokenSource;
         private IDataflowBlock pipelineHead;
         private DiagnosticPipelineConfiguration pipelineConfiguration;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -20,10 +20,11 @@ namespace Microsoft.Diagnostics.EventFlow
     public class DiagnosticPipeline : IDisposable
     {
         private List<IDisposable> inputSubscriptions;
-        private IDisposable batcherTimer;
+        private Timer batcherTimer;
         private List<Task> outputCompletionTasks;
         private bool disposed;
         private bool disposeDependencies;
+        private object disposalLock;
```
xiaomi7732: This is more of a batcherTimerDisposalLock than a general disposalLock, can we rename it to make it explicit?

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
```diff
@@ -182,17 +178,41 @@ public class DiagnosticPipeline : IDisposable
 
             this.disposed = false;
             this.disposeDependencies = disposeDependencies;
+
+            this.batcherTimer = new Timer(
+                (unused) => {
```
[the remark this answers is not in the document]
karolz-ms: Changed.

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/DiagnosticPipeline.cs
karolz-ms: Good suggestion, thanks!

PR #328, opened 2019-08-26
Updated readme to clarify where eventFlowConfig.json should be created

WhitWaldo:
The readme in its current state doesn't indicate where exactly Nuget used to create eventFlowConfig.json at. Rather,  this information is available in the [Service Fabric documentation](https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-diagnostics-event-aggregation-eventflow), so this is just a minor tweak to reflect that for newcomers.


karolz-ms:
Should be fixed now, see https://github.com/Azure/diagnostics-eventflow/commit/6d8aabbfa66b47195281857441df199b2d1d9d4a

Issue #329, opened 2019-09-03
ElasticSearchOutput is using old versions of NEST


acid84:
Hello,

The ElasticSearchoutput have today a dependency to the nuget package NEST that is used for communicate against elasticsearch. (NEST (>= 6.1.0 && < 7.0.0))

If you look on this matrix (https://github.com/elastic/elasticsearch-net#compatibility-matrix) you can see that its not possible to use NEST version 6 against elasticsearch version 7.

This is a very limited factor of eventflow since it basicly are stopping us from upgrading to latest version of elastic search and elastic have end of life on version 6.8 when the release of version 8 is done https://www.elastic.co/support/eol (which is already out in alpha, https://www.elastic.co/guide/en/elasticsearch/reference/master/es-release-notes.html) 

Sorry for many links and thanks for a great tool!

karolz-ms:
@acid84 thanks for reporting the issue and sorry for late reply. This is very helpful. 

How about we publish a new version of the Elasticsearch output (e.g. 2.7.0) that uses NEST 7.x series? We can document that Elasticsearch output 2.6.1 should be used with 6.x ES servers, and 2.7.x should be used with 7.x servers.

karolz-ms:
Also note that ES dropped support for .NET 4.5 and netstandard 1.6 in Elasticsearch.NET/NEST v. 7.x. They do support version 4.6 of the framework and netstandard 2.0. On Eventflow side we support 4.5 or 4.7 (and netstandard 2.0). So we will be making Elasticsearch output version 2.7.x a 4.7/netstandard 2.0 ONLY. Let us know if this is an issue for you.

karolz-ms:
@acid84 Elasticsearch output package version 2.7.0 has been published on NuGet. Please give it a try and let us know how it works for you

PR #331, opened 2019-09-06
Support for Elasticsearch 7

karolz-ms:
Addresses [issue 329](https://github.com/Azure/diagnostics-eventflow/issues/329)


src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -98,7 +98,7 @@ public async Task SendEventsAsync(IReadOnlyCollection<EventData> events, long tr
                 // Note: the NEST client is documented to be thread-safe so it should be OK to just reuse the this.esClient instance
                 // between different SendEventsAsync callbacks.
                 // Reference: https://www.elastic.co/blog/nest-and-elasticsearch-net-1-3
-                IBulkResponse response = await this.connectionData.Client.BulkAsync(request).ConfigureAwait(false);
+                BulkResponse response = await this.connectionData.Client.BulkAsync(request).ConfigureAwait(false);
```
xiaomi7732: nit: Any reason to use class instead of an interface for the response?

karolz-ms: Good question--the reason is Elasticsearch decided to change the interface and use classes instead; `IBulkResponse` is no more ☺️

PR #335, opened 2019-09-19
Set event's Timestamp to DateTimePrecise.UtcNow

popoman:
* Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging
* Microsoft.Diagnostics.EventFlow.Inputs.Trace
      - set event's Timestamp to DateTimePrecise.UtcNow (instead of DateTime.UtcNow). Now (maybe only on .net framework):
            - when there are multiple inputs that internally uses different way of setting Timestamp property there is mismatch of events ordered by Timestamp and their actual sequence
            - there are multiple events with same Timestamp

(Microsoft.Diagnostics.EventFlow.Inputs.PerformanceCounter still uses DateTime.Now.I didn't change it because events rate is limited)


karolz-ms:
Thank you for the change, @popoman, and sorry for reviewing it late. Also thanks to @antymon4o  for review. Will publish shortly

PR #336, opened 2019-09-24
update for .net core 3.0

crocuis:


karolz-ms:
@crocuis sorry for late reply. We'll look into this shortly

karolz-ms:
Support for .NET Core 3.0 has been added by means of https://github.com/Azure/diagnostics-eventflow/pull/338 and other related changes. Because of that this change is no longer necessary. 

However, we appreciate the time you took to look into this issue. Thank you!!

Issue #337, opened 2019-09-24
Net Core 3.0 blocking issue


ksmithRenweb:
Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging.csproj uses Microsoft.Extensions.Logging 2.0.0 which references Microsoft.Extensions.Logging.Abstractions 2.0.0

EventFlowLogger.cs uses Microsoft.Extensions.Logging.Internal.FormattedLogValues

in Microsoft.Extensions.Logging.Abstractions 3.0.0, FormattedLogValues is moved to a new namespace Microsoft.Extensions.Logging.FormattedLogValues and becomes an internal struct.

Due to this namespace move (and the internal scope wouldn't help) you get this exception and stacktrace.
`   System.TypeLoadException
  HResult=0x80131522
  Message=Could not load type 'Microsoft.Extensions.Logging.Internal.FormattedLogValues' from assembly 'Microsoft.Extensions.Logging.Abstractions, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
  Source=Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging
  StackTrace:
   at Microsoft.Diagnostics.EventFlow.Inputs.EventFlowLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.Logger.<Log>g__LoggerLog|12_0[TState](LogLevel logLevel, EventId eventId, ILogger logger, Exception exception, Func`3 formatter, List`1& exceptions, TState& state) in /_/src/Logging/Logging/src/Logger.cs:line 44`

karolz-ms:
Core package 1.6.0 and MicrosoftLogging 1.4.3 should address this. 

Many thanks for investigation and the PR!

ksmithRenweb:
> Core package 1.6.0 and MicrosoftLogging 1.4.3 should address this.
> 
> Many thanks for investigation and the PR!

@karolz-ms anytime.

PR #338, opened 2019-09-24
Replace FormattedLogValues with IReadOnlyList<KeyValuePair<string, ob…

ksmithRenweb:
https://github.com/Azure/diagnostics-eventflow/issues/337


karolz-ms:
Sorry for late reply and thanks for investigating this issue and working out a fix, @ksmithRenweb ! Will look into this shortly

karolz-ms:
The package with this change (Inputs.MicrosoftLogging version 1.4.3) has now been published on nuget.org

Issue #345, opened 2020-01-17
Last Scope.State why is it ToString'd rather than added as a separate Payload


ksmithRenweb:
To me it's a bug, but there has to be a use case.

I've been fighting this for 3 days since I thought it was a bug in my code. Here is my use case.

var logger = provider.GetService<ILogger<...>>();
var disposable = logger.BeginScope(new Dictionary<string, object> {{"Operation ID", Activity.Current.RootId}});
...
logger.LogInformation("Just did something")
...
disiposable.Dispose();

What I am seeing when it gets to Application Insights (I've tracked it to the Log method in EventFlowLogger.) Is that the Scope has the type in it "System.Collections.Generic.Dictionary`2[System.String,System.Object]".

Since my first Scope only has 1 Key Value Pair in it. It's ignored. So I don't get a Operation ID with some Activity Root Id.

The only way it makes sense is if I am using it wrong and I should always have an extra KVP appended to the end of the Scope to avoid this...

ksmithRenweb:
@karolz-ms Tagging you since you updated it to take dictionaries. Vs. only ReadOnlyList's.

ksmithRenweb:
The summary i sent the Lead Architect in my group. Adding it cause i thought it was funny afterwards.

"adding extra properties like this failes
new Dictionary<string, object> {{"Application", "Maintenance Manager Orchestrator"}}
but this would work
new Dictionary<string, object> {{"Application", "Maintenance Manager Orchestrator"}, {"",null}}
 
forehead smack."

karolz-ms:
Interesting, thanks for submitting this issue @ksmithRenweb Good thinking with the `{"", null}` workaround! ☺️ 

I guess we need more tests in this area (Logger + simple/nested scopes). Will look at this more early next week, OK?

ksmithRenweb:
@karolz-ms Thanks! I'll keep an eye out for next week.

karolz-ms:
Just published Microsoft.Logging 1.4.5 package on nuget.org. This should fix it. Feel free to re-open if you run into any problems!

PR #347, opened 2020-01-23
Fix payload generation when Logger scope does not contain formatted message

karolz-ms:
Fixes issue https://github.com/Azure/diagnostics-eventflow/issues/345

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
@@ -116,24 +116,18 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
             // If the data is not structured, it will be omitted
             bool unformattedMessageOmitted = true;
 
-            var structuredSrc = src as IEnumerable<KeyValuePair<string, object>>;
-            if (structuredSrc != null)
+            if (src is IEnumerable<KeyValuePair<string, object>> structuredSrc)
             {
                 unformattedMessageOmitted = false;
-                using (var srcEnumerator = structuredSrc.GetEnumerator())
+                foreach(KeyValuePair<string, object> item in structuredSrc)
                 {
-                    while (srcEnumerator.MoveNext())
+                    if (!OriginalFormatKey.Equals(item.Key, StringComparison.Ordinal))
                     {
-                        var item = srcEnumerator.Current;
-
-                        if (!OriginalFormatKey.Equals(item.Key, StringComparison.Ordinal))
-                        {
-                            action(item);
-                        }
-                        else
-                        {
-                            unformattedMessageOmitted = true;
-                        }
+                        action(item);
+                    }
+                    else
+                    {
+                        unformattedMessageOmitted = true;
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


karolz-ms:
@ksmithRenweb would appreciate your input too (if you have time)

test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/LoggerInputTests.cs
```diff
@@ -358,11 +358,12 @@ public void LoggerIsUsableViaILogger()
                 {
                     var factory = new LoggerFactory();
                     factory.AddEventFlow(diagnosticPipeline);
-                    var logger = new Logger<LoggerInputTests>(factory);
+                    var logger = (ILogger) new Logger<LoggerInputTests>(factory);
```
xiaomi7732: Why not request a clear type like
```csharp
ILogger logger = new Logger<LoggerInputTests>(factory);
```

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
```diff
@@ -102,27 +106,39 @@ private LogLevel ToLogLevel(Extensions.Logging.LogLevel loggerLevel)
...
         {
+            // Special case the "{OriginalFormat}" data key that is used to preserve the message before formatting was applied
+            // This data is omitted from the payload and replaced by formatted message.
+            // See https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
+            const string OriginalFormatKey = "{OriginalFormat}";
+
+            // If the data is not structured, it will be omitted
+            bool unformattedMessageOmitted = true;
+
             var structuredSrc = src as IEnumerable<KeyValuePair<string, object>>;
```
xiaomi7732: How about:
```csharp
if(src is IEnumerable<KeyValuePari<string,object> structuredSrc)
{
...
}
```

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
```diff
@@ -102,27 +106,39 @@ private LogLevel ToLogLevel(Extensions.Logging.LogLevel loggerLevel)
...
             if (structuredSrc != null)
             {
+                unformattedMessageOmitted = false;
                 using (var srcEnumerator = structuredSrc.GetEnumerator())
```
xiaomi7732: Why not just use `foreach` on `IEnumerable<T1,T2>`?

test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/LoggerInputTests.cs
karolz-ms: Makes sense, fixed

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
karolz-ms: Nice, fixed

src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs
karolz-ms: That was refactored from original code that needed to use the enumerator directly. Fixed now!

karolz-ms:
Thanks for the review, @xiaomi7732 Much appreciated

ksmithRenweb:
@karolz-ms I missed the post, but I've looked at the commits. They look good. They also confirm my guess to be pretty close to accurate as to the intent. 

Thanks for getting to this so fast. My PR in my groups sdk's hasn't even finished yet. So I can remove my hacks. We always prefer not to have to maintain such work arounds.

Issue #354, opened 2020-03-05
Missing Logs or Memory Leak in AspNetCore 3.1


timothy-carter:
Using eventflow with the Microsoft ILogger extensions library in aspnetcore 3.1 causes a memory leak when using a Newtonsoft.Json version of 11.0.1 or higher. If Newtonsoft.Json version 10.0.3 is used (as is referenced in the StdOutput library) logs are lost and a PlatformNotSupportedException is thrown.

This has only been observed when using a default log level of Information, but there could be other places where this will happen. When using the Information log level, the [log from aspnetcore](https://github.com/dotnet/aspnetcore/blob/47cc418dec845031f26cc93170ab26c9ad890a07/src/Mvc/Mvc.Core/src/MvcCoreLoggerExtensions.cs#L746-L752) passes a MethodInfo object as one of the values in the state object.

From here, the [EventFlowLogger ](https://github.com/Azure/diagnostics-eventflow/blob/855a93ed9e393a5ee43b7c26465858100c1ba6a5/src/Microsoft.Diagnostics.EventFlow.Inputs.MicrosoftLogging/Logger/EventFlowLogger.cs#L43) is adding that object to the payload.

If using an output such as StdOutput or EventHubOutput, this object gets serialised using JsonConvert.SerializeObject, which attempts to serialise the whole assembly, and memory use goes through the roof.

For a working example of the bug see https://github.com/timothy-carter/aspnetcore/tree/master/EventFlow/MemoryLeak

A related example at https://github.com/timothy-carter/aspnetcore/tree/master/EventFlow/PlatformNotSupportedException uses Newtonsoft.Json version 10.0.3, but this results in logs being lost and a number of PlatformNotSupportedExceptions being thrown.

karolz-ms:
@timothy-carter thank you for reporting this issue and for providing repro examples. It helps a lot and is much appreciated! 

We will try to fix this ASAP, but I need some time to investigate the options. My current idea is to leverage Newtonsoft `JsonSerializerSettings` to special-case types deriving from [MemberInfo](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo?view=netcore-3.1) and just output the fully-qualified name of the member. I would also like to give EventFlow users the ability to customize the `JsonSerializerSettings` so they can change this default behavior if desired.

[...]

PR #355, opened 2020-03-09
Fix issue 354

karolz-ms:
Adds a serialization converter that special-cases all types deriving from reflection library's `MemberInfo`, which will be represented by the member fully-qualified name (e.g. `System.Object.GetHashCode`). This is to avoid serializing the whole reflection graph when using JSON serialization.

Also adds ability to change the JSON serialization settings for `EventHubOutput`, `HttpOutput`, `OmsOutput` (Azure Monitor), and `StdOutput`.

Updates Newtonsoft JSON library dependency to 11.0.2, and, correspondingly, `Microsoft.Configuration.Abstractions` dependency to 2.2.0

Fixes [issue 354](https://github.com/Azure/diagnostics-eventflow/issues/354)


karolz-ms:
Thanks, @xiaomi7732 ! ☺️

PR #365, opened 2020-08-11
Aded Json Converters for IPEndpoint and IPAddress

n-sidorov:
Some libraries (for example, [Orleans](https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Networking/ConnectionLogScope.cs)) may attach logging scopes that contain `IPEndpoint` or `IPAddress` types. 

Both types cannot be serialized with default JSON settings and if EventFlow is configured to read `Microsoft.Extension.Logging` input with those scopes attached it will fail to process those events.

This PR adds simple JSON converters that will call `.ToString()` methods on those types instead of trying to serialize them as Json objects.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/JsonConverters/IPEndpointConverter.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/JsonConverters/IPEndpointConverter.cs
@@ -1,0 +1,38 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Net;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/JsonConverters/IPEndpointConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Microsoft.Diagnostics.EventFlow.JsonConverters
+{
+    internal class IPEndpointConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+#if NET451 || NETSTANDARD1_6
+            return typeof(EndPoint).GetTypeInfo().IsAssignableFrom(objectType);
+#else
+            return typeof(EndPoint).IsAssignableFrom(objectType);
```
n-sidorov, line 22: Note:
There are 3 implementations of `EndPoint` and all of them override `.ToString()`.

PR #366, opened 2020-08-24
Add Evaluator to check for missing property

ksmithRenweb:
We need a way to filter based off properties not being on an event. 

There didn't appear to be a way to do that yet within the existing evaluators.


karolz-ms:
@ksmithRenweb thank you for submitting the PR!! I think it is a very good idea!

However, I would not implement this concept as `EventPropertyExpressionEvaluator`, since here is no "property expression" to evaluate--we are checking whether the property exists at all. The proposed implementation will be quite confusing to use: `transactionId is 1234` actually means "pass events that do not have `transactionId` property" 😮 ☺️ 

Also, your attempt will not work because [the grammar](https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.FilterParserGenerator/FilterParser.peg) does not allow it (there is no "`is`" operator that can go between property name and the value--see lines 30-38. So you will get a parsing error when you try to use it.

Instead, I would suggest that we add two new evaluators: `hasproperty` and `hasnoproperty`, so your filtering expression could be something like `hasproperty foo && timestamp > "20200825"` or `hasnoproperty bar`

This is a more complex change, and I am not sure how much time you have on hand to work on this, so there is several ways we can go about it:

- you can attempt what I sketched above
- we can collaborate (e.g. I can modify the grammar and you do the implementation)
- I can do the implementation and you do the code review and evaluate whether the change suits your needs

Let me know what works best for you. Thanks again for the idea and the prototype.

ksmithRenweb:
@karolz-ms Please review. I either just got it right, or i need to understand peg files more. ;) 
I don't have time to test it locally, been working a production issue all day. I needed a break, so I looked at this.

PR #376, opened 2020-12-30
Fix race condition in EventFlowSubject

karolz-ms:
While doing some other work I noticed we have a race condition in `EventFlowSubject`. Basically, if `OnNext()` is preempted just between the `observers` check and setting `notifyingObservers` to true (line 53 in the old code), the `Shutdown()` method running on separate thread will complete, including calling `OnCompleted()` on the observers. That will then be followed by `OnNext() (catching up after resumption). So `OnCompleted()` will be followed by `OnNext()`, which is not what we want (breaks the contract).

This change fixes this problem.

The `FirstChanceExceptionCounter` change is unrelated; it is just a small change to fix spurious "unused variable" warning when targeting .NET Standard 1.6.

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventFlowSubject.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventFlowSubject.cs
@@ -15,10 +15,27 @@ namespace Microsoft.Diagnostics.EventFlow
 {
     public class EventFlowSubject<SubjectType> : IObservable<SubjectType>, IObserver<SubjectType>, IDisposable, IItemWithLabels
     {
-        private readonly object lockObject = new object();
-        private volatile ImmutableList<IObserver<SubjectType>> observers = ImmutableList<IObserver<SubjectType>>.Empty;
-        private volatile bool notifyingObservers = false;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventFlowSubject.cs
```diff
@@ -47,24 +64,24 @@ public void OnError(Exception error)
 
         public void OnNext(SubjectType value)
```
yantang-msft, line 65: ```OnNext``` can be run in multiple threads right? In this case, I can see that the state of ```notifyingObservers``` is just not correctly maintained in the original implementation, even without the ```ShutDown()``` method. E.g., thread 1 at line 57, thread 2 at line 67, ```notifyingObservers``` is set to false but it's actually still notfiying the observers.
The new implementation looks much better.

yantang-msft (approved):

src/Microsoft.Diagnostics.EventFlow.Core/Implementations/EventFlowSubject.cs
karolz-ms: Yes, that is right, good point!

Issue #377, opened 2021-01-05
EventFlow not working with .NET 5


ChrisRobbers:
Our app has been using EventFlow with the Microsoft.Extensions.Logging input and Azure EventHub output for many months, but when I tried upgrading the app to .NET 5 I noticed it stopped sending to our EventHub. Changing it back to .NET Core 3.1 caused it to start working again. I don't know if this is something specific to our configuration or something in general with EventFlow. Has this been tested against .NET 5 in general?

Here is our eventFlowConfig.json:

```
{
  "inputs": [
    {
      "type": "Microsoft.Extensions.Logging"
    }
  ],
  "outputs": [
    {
      "type": "EventHub",
      "eventHubName": "",
      "connectionString": "",
      "filters": [
        {
          "type": "drop",
          "include": "Level == Verbose"
        }
      ]
    }
  ],
  "settings": {
    "maxEventBatchSize": "100",
    "maxBatchDelayMsec": "5000"
  },
  "schemaVersion": "2016-08-11"
}
```

karolz-ms:
@ChrisRobbers we are just getting to it. Thanks for the report, we'll look ASAP

yantang-msft:
@ChrisRobbers Thanks for reporting this issue. I reproduced it and found this is because of this bug: https://github.com/Azure/azure-sdk-for-net/issues/13899
We will update our dependency on EventHub to the latest version (from 1.0.1 to 4.3.1, yes it's quite old) and this issue will go away.

karolz-ms:
Many thanks @yantang-msft for investigation and the fix!

@ChrisRobbers a new version of the EventHub output with the fix (1.7.0) has been published on nuget.org. Please let us know if there are any issues with it, otherwise enjoy ☺️

PR #379, opened 2021-01-16
Upgrade to latest EventHub nuget package so the EventHub output work with .Net 5

yantang-msft:
This is to address this issue: https://github.com/Azure/diagnostics-eventflow/issues/377

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Microsoft.Diagnostics.EventFlow.Outputs.EventHub.csproj
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Microsoft.Diagnostics.EventFlow.Outputs.EventHub.csproj
@@ -6,7 +6,7 @@
     <Authors>Microsoft</Authors>
     <TargetFrameworks>net452;netstandard1.6;netstandard2.0</TargetFrameworks>
     <AssemblyName>Microsoft.Diagnostics.EventFlow.Outputs.EventHub</AssemblyName>
-    <VersionPrefix>1.6.0</VersionPrefix>
+    <VersionPrefix>1.7.0</VersionPrefix>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <PackageId>Microsoft.Diagnostics.EventFlow.Outputs.EventHub</PackageId>
     <PackageTags>Microsoft;Diagnostics;EventFlow;Outputs;Event Hub;Event Hubs</PackageTags>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Microsoft.Diagnostics.EventFlow.Outputs.EventHub.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>Provides an output implementation that sends diagnostics data to Azure Event Hubs.</Description>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <Authors>Microsoft</Authors>
-    <TargetFrameworks>net452;netstandard1.6;netstandard2.0</TargetFrameworks>
```
karolz-ms, line 7: We could also keep producing net452 binary (upgrade the dependency to ver 1.0.3), and refer to version 4.3.1 for netstandard 1.6 and 2.0. .NET 4.5.2 is still supported so maybe it is too early to discontinue that...

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Microsoft.Diagnostics.EventFlow.Outputs.EventHub.csproj
```diff
@@ -4,7 +4,7 @@
...
+    <TargetFrameworks>netstandard1.6;netstandard2.0</TargetFrameworks>
     <AssemblyName>Microsoft.Diagnostics.EventFlow.Outputs.EventHub</AssemblyName>
     <VersionPrefix>1.6.0</VersionPrefix>
```
karolz-ms, line 9: You want to increment the package version to something like 1.7.0

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/Microsoft.Diagnostics.EventFlow.Outputs.Tests.csproj
```diff
@@ -20,15 +20,15 @@
...
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net471' ">
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.14.0" />
-    <PackageReference Include="Microsoft.Azure.EventHubs" Version="1.0.1" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />    
+    <PackageReference Include="Microsoft.Azure.EventHubs" Version="4.3.1" />
```
karolz-ms, line 30: The explicit reference to Microsoft.Azure.EventHubs can probably be dropped from this project.

karolz-ms (approved):
Thanks, Yanming!

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Microsoft.Diagnostics.EventFlow.Outputs.EventHub.csproj
yantang-msft: That's a good point

PR #382, opened 2021-03-23
[WIP] (Question on customizing ETW input) - ETW/MsgPack input based off StandardTraceEventSession input

maxgolov:
Greetings!

I would like to get some feedback and suggestions on how to implement the decoding of ETW events that are encoded in MsgPack. In some flows in Azure flows we utilize this encoding (MsgPack), which is neither Manifest nor TraceLogging Dynamic, but rather carries MsgPack-encoded payload that is generated by the application at runtime.

I implemented the payload decoder based off standard `EtwInput.cs`.

It required a few modifications:

- configuration flag **encoding:  MsgPack** that allows to specify that the incoming events should be decoded as MsgPack. Example config:

```json
{
  "inputs": [
    {
        "type": "ETW",
        "sessionNamePrefix": "OpenTelemetry",
        "cleanupOldSessions": true,
        "reuseExistingSession": true,
        "encoding":  "MsgPack",
        "providers": [
            {
                "providerName": "OpenTelemetry-ETW-MsgPack"
            }
        ]
    }
  ],
    "outputs": [
        {
            "type": "StdOutput"
        }
    ],
  "schemaVersion": "2016-08-11"
}
```

- had to change the `Process()` method signature to `public virtual void Process(Action<EventData> onEvent)`, in order to allow for a custom decoder implementation in `MsgPackTraceEventSession`. I am not sure about performance implications of that.

- changed a few class members from `private` to `protected`, to allow the `MsgPackTraceEventSession` inherit from the base `StandardTraceEventSession`

Eventual goal is to allow OpenTelemetry **C++ SDK** to emit ETW events using either TraceLoggingDynamic (StandardTraceEventSession) or ETW/MsgPack (MsgPackTraceEventSession), for easy integration between OpenTelemetry **C++ SDK** running on Windows with EventFlow, using EventFlow as a debugging / diagnostics tool, as well as forwarder to Azure Monitor, Application Insights, plain stdout, etc.

Is there a better / alternate approach available, maybe some extension point that would've allowed me to register a custom callback / decoder? Activity related to https://github.com/open-telemetry/opentelemetry-cpp/issues/326 and PR https://github.com/open-telemetry/opentelemetry-cpp/pull/519

Thank you!

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
@@ -1,0 +1,166 @@
+﻿using Microsoft.Diagnostics.Tracing;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+using MsgPack.Serialization;
+using Microsoft.Diagnostics.EventFlow.Utilities.Etw;
+
+namespace Microsoft.Diagnostics.EventFlow.Inputs
+{
+    internal class MsgPackTraceEventSession : StandardTraceEventSession
+    {
+
+        /// <summary>
+        /// MsgPackTraceEventSession allows to decode MsgPack-encoded binary events.
+        /// </summary>
+        /// <param name="sessionNamePrefix">See @base</param>
+        /// <param name="cleanupOldSessions">See @base</param>
+        /// <param name="reuseExisting">See @base</param>
+        /// <param name="healthReporter">See @base</param>
+        public MsgPackTraceEventSession(string sessionNamePrefix, bool cleanupOldSessions, bool reuseExisting, IHealthReporter healthReporter) :
+            base(sessionNamePrefix, cleanupOldSessions, reuseExisting, healthReporter)
+        {
+        }
+
+        /// <summary>
+        /// Convert TraceEvent in MsgPack format to EventData
+        /// </summary>
+        /// <param name="data"></param>
+        /// <param name="reporter"></param>
+        /// <returns></returns>
+        internal static EventData DecodeMessagePack(TraceEvent data, IHealthReporter reporter)
+        {
+            EventData eventData = new EventData
+            {
+                ProviderName = data.ProviderName,
+                Timestamp = data.TimeStamp,
+                Level = (LogLevel)data.Level,
+                Keywords = (long)data.Keywords,
+            };
+
+            if (data.EventDataLength==0)
+            {
+                reporter.ReportProblem("MsgPack decoding failed: empty payload!");
+                return eventData;
+            }
+
+            byte[] managedArray = new byte[data.EventDataLength];
+            Marshal.Copy(data.DataStart, managedArray, 0, data.EventDataLength);
+
+            try
+            {
+                var obj = MessagePackSerializer.UnpackMessagePackObject(managedArray);
+                if (obj.IsDictionary)
+                {
+                    foreach (var kv in obj.AsDictionary())
+                    {
+                        var key = kv.Key.AsString();
+                        object val = null;
+                        switch (Type.GetTypeCode(kv.Value.UnderlyingType))
+                        {
+                            case TypeCode.String:
+                                val = kv.Value.AsString();
+                                break;
+
+                            case TypeCode.Boolean:
+                                val = kv.Value.AsBoolean();
+                                break;
+
+                            case TypeCode.Decimal:
+                                val = kv.Value.AsInt64();
+                                break;
+
+                            case TypeCode.Int16:
+                                val = kv.Value.AsInt16();
+                                break;
+
+                            case TypeCode.Int32:
+                                val = kv.Value.AsInt32();
+                                break;
+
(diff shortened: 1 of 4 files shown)


karolz-ms:
@maxgolov what you did makes a lot of sense, thank you for your contribution! I do not think there is a significantly better way/extension point to accomplish what you are trying to do.

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
```diff
@@ -0,0 +1,166 @@
...
+                Keywords = (long)data.Keywords,
+            };
+
+            if (data.EventDataLength==0)
+            {
+                reporter.ReportProblem("MsgPack decoding failed: empty payload!");
+                return eventData;
+            }
+
+            byte[] managedArray = new byte[data.EventDataLength];
```
karolz-ms, line 50: Will this buffer be allocated every time a new event comes? Can/should we reduce allocation by using something like `ArrayPool<T>`?

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
```diff
@@ -0,0 +1,166 @@
...
+                }
+                else
+                {
+                    eventData.Payload["Message"] = obj.ToString();
+                }
+
+            }
+            catch (ArgumentNullException ex)
+            {
+                reporter.ReportProblem("MsgPack decoding failed: " + ex.ToString());
```
karolz-ms, line 124: See comment above about reporting error vs warning if a single event fails to decode

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
```diff
@@ -0,0 +1,166 @@
...
+                        eventData.Payload[key] = val;
+                    }
+                }
+                else
+                {
+                    eventData.Payload["Message"] = obj.ToString();
+                }
+
+            }
+            catch (ArgumentNullException ex)
```
karolz-ms, line 122: What is the reason behind catching only `ArgumentNullException`? Are there other exception types that can be raised during MessagePack parsing that we could/should suppress?

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
karolz-ms: `Span<T>` will probably be even better if `EventDataLength` is not unreasonably large and if `UnpackMessagePackObject` can work with spans (and if the latter is not true, we might consider contributing a patch ☺️ )

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
```diff
@@ -0,0 +1,166 @@
+﻿using Microsoft.Diagnostics.Tracing;
```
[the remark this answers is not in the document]
maxgolov: Yes. I'll add the same copyright as the rest of code, as I'm acting as Microsoft Employee w.r.t. this commit :) I am not 100% sure if this input adapter needs to live in the mainline, or moved elsewhere. It would be great if it is in the mainline, assuming that adding dependency on `MsgPack/CLI` package is tolerable.

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                ProviderName = data.ProviderName,
+                Timestamp = data.TimeStamp,
+                Level = (LogLevel)data.Level,
+                Keywords = (long)data.Keywords,
+            };
+
+            if (data.EventDataLength==0)
+            {
+                reporter.ReportProblem("MsgPack decoding failed: empty payload!");
```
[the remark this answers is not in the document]
maxgolov: Agreed. I'll fix this.

src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs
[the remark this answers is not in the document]
karolz-ms: Yes, taking a dependency on `MsgPack/CLI` is fine, we have other parts of EventFlow that depend on Apache-licensed NuGet packages, Elasticsearch for example.

maxgolov:
Apologies for delays with this. I'll get back to this PR in May 2021.

karolz-ms:
@maxgolov I assume the plans have changed ☺️ LMK if you want me to keep this PR, or close, either is fine with me.

maxgolov:
@karolz-ms - Karol, apologies. Indeed, the plans have changed. We got ETW working successfully for what we wanted it work in OpenTelemetry. But further work on EventFlow is on a back-burner for me, outside of my ownership domain right now, as I moved to a different project. Closing it.

PR #383, opened 2021-04-08
Update readme

dtPaTh:
Referencing Dynatrace output extension


karolz-ms (approved):
Thank you, @dtPaTh for this change and for adding EventFlow output for Dynatrace 🚀

Issue #385, opened 2021-04-22
Duplicate dependencies in elastic search output


arwyl:
https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs

As I can see code blocks from lines 154-165 and 167-178 are identical

so dependencies are yielded twice

karolz-ms:
😊  Sorry about that @arwyl Good catch! We'll try to fix it ASAP

karolz-ms:
Re-opening till new NuGets are published

karolz-ms:
Elasticsearch output package 2.7.3 has been published to nuget.org

PR #386, opened 2021-05-07
Bug fix and dependency update

karolz-ms:


src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -164,19 +164,6 @@ private IEnumerable<IBulkOperation> GetCreateOperationsForEvent(EventData eventD
                 }
             }
 
-            if (eventData.TryGetMetadata(DependencyData.DependencyMetadataKind, out metadataSet))
```
yantang-msft, line 167: What's the bug? Is CreateDependencyOperation() still needed then?

karolz-ms: Duplicate code, see https://github.com/Azure/diagnostics-eventflow/issues/385

karolz-ms:
@yantang-msft FYI build is failing because we had to switch to a new agent pool and that agent pool cannot use our current method of publishing build artifacts 😣 Need to fix that...

yantang-msft (approved):

karolz-ms:
Builds are now working

PR #387, opened 2021-09-22
Updated EventHub output to use Azure.* packages

n-sidorov:
This PR updates EventHub output package to use `Azure.*` libraries and adds support for using "Azure Identity" for EventHub authentication (instead of Connection Strings).

BREAKING CHANGES: because `Azure.Messaging.EventHubs` only targets `netstandard2.0` this new package also targets only `nestandard2.0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -852,7 +852,11 @@ This output writes data to a webserver using diffent encoding methods (Json or J
 ### Event Hub
 *Nuget Package*: [**Microsoft.Diagnostics.EventFlow.Outputs.EventHub**](https://www.nuget.org/packages/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/)
 
-This output writes data to the [Azure Event Hub](https://azure.microsoft.com/en-us/documentation/articles/event-hubs-overview/). Here is an example showing all possible settings:
+> **BREAKING CHANGE**: Starting from version `2.0.0` this package targets _only_ `netstandard2.0`.
+
+This output writes data to the [Azure Event Hub](https://azure.microsoft.com/en-us/documentation/articles/event-hubs-overview/).
+
+Here is an example showing configuration using connection string:
 ```jsonc
 {
     "type": "EventHub",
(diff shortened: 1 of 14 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
```diff
@@ -187,23 +182,53 @@ private async void CleanUpAsync()
...
+
+                if (string.IsNullOrWhiteSpace(this.outputConfiguration.FullyQualifiedNamespace))
+                {
+                    var emptyNamespaceMsg = $"{nameof(EventHubOutput)}: Event Hub namespace must not be empty when using Azure Identity. It can be specified in the '{nameof(EventHubOutputConfiguration.FullyQualifiedNamespace)}' configuration parameter";
+                    healthReporter.ReportProblem(emptyNamespaceMsg, EventFlowContextIdentifiers.Configuration);
+                    throw new Exception(emptyNamespaceMsg);
+                }
+
+                return new EventHubClientImpl(
+                    new EventHubProducerClient(this.outputConfiguration.FullyQualifiedNamespace, this.eventHubName, new DefaultAzureCredential())
```
n-sidorov, line 202: So, this is kind of sketchy, but since EventFlow configured via "json file" I am not sure we need to expose `ITokenCredential`.

README.md
```diff
@@ -852,21 +852,42 @@ This output writes data to a webserver using diffent encoding methods (Json or J
 ### Event Hub
 *Nuget Package*: [**Microsoft.Diagnostics.EventFlow.Outputs.EventHub**](https://www.nuget.org/packages/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/)
 
-This output writes data to the [Azure Event Hub](https://azure.microsoft.com/en-us/documentation/articles/event-hubs-overview/). Here is an example showing all possible settings:
+> **BREAKING CHANGE**: Starting from version `2.0.0` this package targets _only_ `netstandard2.0`.
```
karolz-ms, line 855: There is a "Platform Support" section at the end of the document; I think the information about it should go there. Namely, that version 1.x of the output package support 4.5.1, 4.6.1 and .NET Core frameworks, but version 2.x supports only 4.6.1 and .NET Core/.NET 5 frameworks

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventDataExtensions.cs
```diff
@@ -49,7 +47,7 @@ public static MessagingEventData ToMessagingEventData(this EventData eventData,
                 : ConvertToShoeboxTrace(eventData);
 
             // If this turns out to consume significant CPU time, we could serialize the object "manually".
-            // See https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs for an example.
+            // See https://github.com/dotnet/runtime/blob/0ea3b7d8cf957508a94061b8e921b28d6f6becd6/src/libraries/Microsoft.Extensions.Logging.EventSource/src/EventSourceLogger.cs for an example.
```
karolz-ms, line 50:
```suggestion
            // See https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.EventSource/src/EventSourceLogger.cs for an example.
```

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
karolz-ms: The design principle with EventFlow is that JSON configuration is for common use cases, and programmatic configuration is for advanced use cases. It is OK for certain configurations to be doable only through programmatic configuration, but the opposite should never be the case.

Consider adding `Credential` property to `EventHubOutputConfiguration` and using it here.

How common is it to use a non-default credential?

karolz-ms (commented):
Looks great overall, thank you, just a couple questions/suggestions

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventDataExtensions.cs
n-sidorov: Why though? This is permalink that will 
"always" point to a code that is references in comment. Having `main` will point to "latest" which may not have code that is references in the comment?

karolz-ms: You are right--it is more important to refer to code that shows the example than to have a reference to current code.

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
n-sidorov: It's generally pointless to have "DefaultAzureCredential" in Azure since in Azure you only need "ManagedAzureCredential".

But locally, "DefaultAzureCredential" is a way to go since it includes everything (your Visual Studio credentials, Azure CLI, Managed Identity etc).

I think exposing _optional_ `ITokenCredential` is a good idea.

n-sidorov: Updated to include `TokenCredential`. Also, based on docs, having `DefaultTokenCredential` should be fine: https://docs.microsoft.com/en-us/dotnet/api/overview/azure/identity-readme#defaultazurecredential

README.md
n-sidorov: Updated "Plarform Support".

n-sidorov:
@karolz-ms So, I've found that tests that checking that Input/Output can be created were busted for `net461` target in VS. It looks like there were dependency resolution issues. So I've updated targeting for some of the libs to make it work. If you think it's out of scope of this PR, then I will revert it.

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/EventHubOutput.cs
karolz-ms: great, thanks

karolz-ms (approved):

karolz-ms:
> @karolz-ms So, I've found that tests that checking that Input/Output can be created were busted for `net461` target in VS. It looks like there were dependency resolution issues. So I've updated targeting for some of the libs to make it work. If you think it's out of scope of this PR, then I will revert it.

No, it is great, thank for fixing this!

Issue #394, opened 2021-11-11
Need to set proxy for ElasticSearch output


srothkin:
Our on-prem environment requires use of a proxy to get to our AWS-hosted Elastic server. By default, NEST (at least under .Net Framework 4.8) doesn't use the system default proxy configuration. And EventFlow Elastic output doesn't expose the underlying NEST proxy configuration settings.

Suggestion - add the various proxy settings in https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/configuration-options.html to EventFlow's Elastic output configuration and use those when creating the Elastic client.

karolz-ms:
@srothkin thank you for the suggestion. Looks like the API surface for the ES connection options is pretty large. Also it is likely it will be changing from one ES version to the next. For these reasons I am reluctant to attempt to reflect ES connection options into EventFlow configuration schema.

Instead I am thinking we could provide a callback that would allow Elasticsearch output user to customize the `ConnectionOptions` object that is used by the output to connect to Elasticsearch. The only drawback is that this capability would be available via code. In other words, ES output, together with the rest of EventFlow pipeline would have to be created completely programmatically. At least that is the case today with the current diagnostic pipeline design. Let us know if this would be an acceptable solution.

srothkin:
No, that's not acceptable. I'd like it to be kept simple - an optional proxy URL in the Elastic Output config JSON - if set, the Elastic Output adds that to the configuration object before creating the Elastic Client. 
I don't really expect that the proxy url option in the underlying NEST config is going to change from version to version unless they have a larger change that would break other things as well.

karolz-ms:
OK, sorry, I misunderstood your suggestion--thought you were asking to support all ES connection configuration properties. 

Can you confirm that setting the proxy address, username, and password will be sufficient for your use case?

srothkin:
Already tested/verified that with Internet Explorer proxy settings (Windows Server 2012 R2).

karolz-ms:
@srothkin Elasticsearch output package 2.7.4 has been published to NuGet. Could you give it a test and report if it solves this issue?

srothkin:
It does, thank you.

PR #395, opened 2021-11-13
Add connection proxy settings to Elasticsearch output

karolz-ms:
Implements https://github.com/Azure/diagnostics-eventflow/issues/394

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/Configuration/ElasticSearchProxyConfiguration.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/Configuration/ElasticSearchProxyConfiguration.cs
@@ -1,0 +1,29 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.EventFlow.Configuration
+{
+    public class ElasticSearchProxyConfiguration
+    {
+        public string Uri { get; set; }
+        public string UserName { get; set; }
+        public string UserPassword { get; set; }
+
+        internal ElasticSearchProxyConfiguration DeepClone()
+        {
+            var other = new ElasticSearchProxyConfiguration();
+            other.Uri = this.Uri;
+            other.UserName = this.UserName;
+            other.UserPassword = this.UserPassword;
+            return other;
+        }
+    }
+}
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
@@ -309,6 +309,24 @@ namespace Microsoft.Diagnostics.EventFlow.Outputs
             {
                 connectionSettings = connectionSettings.BasicAuthentication(userName, password);
             }
+            if (!string.IsNullOrWhiteSpace(esOutputConfiguration.Proxy.Uri))
+            {
+                Uri proxyUri = null;
+                try
+                {
+                    proxyUri = new Uri(esOutputConfiguration.Proxy.Uri);
+                }
+                catch { }
+                if (proxyUri == null)
+                {
+                    var errorMessage = $"{nameof(ElasticSearchOutput)}: proxy URI setting value '{esOutputConfiguration.Proxy.Uri}' is not valid";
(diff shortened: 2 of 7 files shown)


src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/Configuration/ElasticSearchProxyConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.EventFlow.Configuration
+{
+    public class ElasticSearchProxyConfiguration
+    {
+        public string Uri { get; set; }
```
xiaomi7732, line 16: Is there a reason not to use type of `Uri` or `Uri?` directly?

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -309,6 +309,24 @@ private void Initialize(ElasticSearchOutputConfiguration esOutputConfiguration)
             {
                 connectionSettings = connectionSettings.BasicAuthentication(userName, password);
             }
+            if (!string.IsNullOrWhiteSpace(esOutputConfiguration.Proxy.Uri))
+            {
+                Uri proxyUri = null;
```
xiaomi7732, line 314: How about:

```csharp
if(Uri.TryParse(esOutputConfiguration.Proxy.Uri, out Uri proxyUri)
{
    connectionSettings.Proxy(proxyUri, ..);
}
else
{
    var errorMessage = $"{nameof(ElasticSearchOutput)}: proxy URI setting value '{esOutputConfiguration.Proxy.Uri}' is not valid";
    healthReporter.ReportWarning(errorMessage, EventFlowContextIdentifiers.Configuration);
}
```

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
```diff
@@ -43,7 +43,13 @@ public void VerifyAllConfigOptionsMappedToConfigObject()
                                     ["type"] = "date_nanos"
                                 }
                             }
-                        }
+                        },
+                        ["proxy"] = new Dictionary<string, object>
```
xiaomi7732, line 47: It is great that we could cover this case here. I am just wondering, is it common for a proxy to be provided without username and password? What do you think to add some unit tests to cover that when userName or password being null or empty?

xiaomi7732 (changes requested):
Generally LGTM. Just a few questions to satisfy my curiosity. Mind give some explanations? Thanks!

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
srothkin: For our on-prem environment there's no username/password for the proxy. And the proxy is plain http. The proxy is being used to limit which internet sites can be accessed from our production servers.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
karolz-ms: Great suggestion, will do that

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/Configuration/ElasticSearchProxyConfiguration.cs
karolz-ms: It is a string in order to avoid dealing with malformed URIs at deserialization time.

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
karolz-ms: Like @srothkin said, empty/null username and passwords are valid use cases.

xiaomi7732: @karolz-ms Your call. But what I meant was to have unit test cover when userName/password is null and the code keeps working. In other words, probably no exception when userName or password is empty.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/Configuration/ElasticSearchProxyConfiguration.cs
xiaomi7732: I see. That makes sense.

xiaomi7732 (approved):
Sign off.

Issue #396, opened 2021-11-17
Shall we remove unused `Microsoft.Diagnostics.EventFlow.TableStorage`?


xiaomi7732:
Hi everyone,

After touch base with @karolz-ms, we reached an agreement to remove `Microsoft.Diagnostics.EventFlow.TableStorage` due to lack of usability, and in the sake of having a reduced attacking surface for vulnerability.

But we want to hear everyone's opinion, specially, ones that vote against the decision, just in case there's some hidden scenarios that we should've realized. Please comment below if you think there's a scenario for you and the project is worth keeping.

**Details**
The package was developed at the beginning of the project but never since released a NuGet package for it. And there's no inquiries about it either.

karolz-ms:
Resolved via https://github.com/Azure/diagnostics-eventflow/pull/398 -- we did remove it

PR #397, opened 2021-12-08
Wipe out table storage output

xiaomi7732:
Addresses #396


xiaomi7732:
I don't understand the conflicts. Let me double check.

PR #398, opened 2021-12-08
Wipe out table storage output

xiaomi7732:
Address #397


karolz-ms (approved):
Thank you @xiaomi7732 🤗

Issue #416, opened 2023-02-15
Add ability to configure index format string


BuddhaBuddy1:
Daily indexes are too big for our situation. We would like to have monthly ones, but there is no way to configure the format string. It is hard coded to 'yyyy.MM.dd'.

I would like to be able to configure the index format string.

BuddhaBuddy1:
I have submitted [PR 417](https://github.com/Azure/diagnostics-eventflow/pull/417) to implement this feature.

karolz-ms:
This is done

PR #417, opened 2023-02-15
Add ability to configure indexFormat

BuddhaBuddy1:
This is to implement [issue 416](https://github.com/Azure/diagnostics-eventflow/issues/416).


BuddhaBuddy1:
I may have jumped the gun here. I may need to add WeekOfYear or QuarterOfYear to the formatting options, but apparently C# string.Format() doesn't support that. /sigh

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -378,10 +379,25 @@ private async Task EnsureIndexExists(string indexName, ElasticClient esClient)
...
+                dateString = string.Format($"{{0:{connectionData.Configuration.IndexFormat}}}", DateTimeOffset.UtcNow);
+            }
+            catch (Exception ex)
+            {
+                string errorMessage = $"{nameof(ElasticSearchOutput)}: indexFormat could not parse as date: {ex.Message}{Environment.NewLine}" +
+                                      $"Invalid IndexFormat value: {connectionData.Configuration.IndexFormat}{Environment.NewLine}" +
+                                      $"Falling back to default value: {ElasticSearchOutputConfiguration.DefaultIndexFormat}";
+                // Changing in connectionData.Configuration so we don't report this every log message
+                connectionData.Configuration.IndexFormat = ElasticSearchOutputConfiguration.DefaultIndexFormat;
+                healthReporter.ReportProblem(errorMessage);
```
karolz-ms, line 396: I would use a warning here

karolz-ms (approved):
Looks great! Thank you, @BuddhaBuddy1. Just one small suggestion

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
BuddhaBuddy1: I have also added new format strings to enable Quarter of Year, Week of Month, and WeekOfYear. Can I add that update at the same time, or should it be separate? It is quite a bit more complex than the previous update, so I can see there being more objections to it that this first PR.

karolz-ms: Typically, with a PR, you would just add more commits to the repository/branch that you used to create the PR. These commits will then be "included" in the PR.  For this PR it would be `BuddhaBuddy1/diagnostics-eventflow:master` repository/branch.

karolz-ms: Also, the new format strings sound very much related to the rest of the enhancement that you are making, so please, go ahead and add them and I will re-review, no problem, and thank you!

BuddhaBuddy1: Changed to warning, new formatting added. If anyone knows how to make the readme changes prettier, feel free. :-)

BuddhaBuddy1: I feel like I made a mistake by merging my changes into my fork's master first, then doing a PR on that. Should I have left it in the branch?

karolz-ms: It is not a huge deal, but the typical workflow that I use is that you create a separate branch for the PR. Then you can use that branch to add more commits to it and/or decide what other changes to merge into that branch. It isolates the work you are doing on the PR from your other work, and from changes that come from upstream repository too.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -378,11 +382,77 @@ private async Task EnsureIndexExists(string indexName, ElasticClient esClient)
...
+                dateString = string.Format($"{{0:{connectionData.Configuration.IndexFormat}}}", DateTimeOffset.UtcNow);
+            }
+
+            return connectionData.Configuration.IndexNamePrefix + dateString;
+        }
+
+        public Func<DateTimeOffset> Now { get; set; } = DefaultNow;
+        public static DateTimeOffset DefaultNow() { return DateTimeOffset.UtcNow; }
+
+        protected internal string ConvertIndexFormat(string dateString)
```
karolz-ms, line 409: I am wondering how fast this method is. .NET has become quite efficient these days, but still, this will be called for every batch of events, so it might be beneficial to optimize things a bit. A few suggestions below.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -378,11 +382,77 @@ private async Task EnsureIndexExists(string indexName, ElasticClient esClient)
...
+        {
+            var index = dateString.IndexOf('q');
```
karolz-ms, line 411: Noticed that this method, and the methods it calls, call into `this.Now()` in a several places. Consider getting the "now" timestamp just once and pass it around by (read-only) reference. This will also avoid (arguably unlikely) cases where these timestamps vary enough to affect the index name.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -378,11 +382,77 @@ private async Task EnsureIndexExists(string indexName, ElasticClient esClient)
...
+            if (index > -1)
+                dateString = dateString.Substring(0, index) + GetQuarterOfYear() + dateString.Substring(index + 1);
+            index = dateString.IndexOf('w');
```
karolz-ms, line 414: Since EventFlow does not support dynamic reconfiguration, you could examine the `dateString` in the constructor of the output and determine what kind of treatment you want to use to get the final `dateString` that you pass to `string.Format`. This would avoid parsing the `dateString` for every batch of events.

 For all special formatting characters that you are introducing here the treatment for determining what the effective `dateString` should be is basically the same:
1. A prefix (does not change from event batch to the next)
2. A string produced by a method (`GetQuarterOfYear` etc) with a single input (the ""now" timestamp) i.e. `Func<DateTimeOffset, String>`
3. A suffix (does not change from event batch to the next)
4. Combine 1 + 2 + 3 to get the effective `dateTime` string.

In fact, even if the user does not use quarter/week of month/week of year, the above generalization still works with prefix=original `dateString`, the method = null, and the suffix = empty string. 

Of course this works if you do not want to support more than one of your special formatting characters ☺️ (e.g. no quarter AND week of year in the same formatting string). But I assume this is the intention here, and we could just document it.

Not sure if all this makes things faster enough to make the additional code complexity worthwhile, so feel free to skip this suggestion.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -378,11 +382,77 @@ private async Task EnsureIndexExists(string indexName, ElasticClient esClient)
...
+        }
+
+        private string GetQuarterOfYear()
+        {
+            return "q"+ (((Now().Month - 1) / 3) + 1);
+        }
+
+        private string GetWeekOfMonth()
+        {
+            DateTimeOffset date = Now();
```
karolz-ms, line 441: This and the next variable do not seem to be used to compute the answer?

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
```diff
@@ -86,9 +88,38 @@ public void VerifyAllConfigOptionsMappedToConfigObject()
...
+                Assert.Equal($"myprefix-{DateTimeOffset.UtcNow:yyyy.MM}", eso.GetIndexName(eso.connectionData));
             }
         }
 
+        [Theory]
+        [InlineData(2023, 1, 1, "yyyy.q", "2023.q1")]
+        [InlineData(2023, 3, 1, "yyyy.q", "2023.q1")]
+        [InlineData(2023, 4, 1, "yyyy.q", "2023.q2")]
+        [InlineData(2023, 1, 1, "yyyy.MM.w", "2023.01.w1")]
+        [InlineData(2023, 2, 1, "yyyy.MM.w", "2023.02.w1")]
```
karolz-ms, line 102: Please add a test for week-of-month that is not the first week ☺️

karolz-ms (changes requested):

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
BuddhaBuddy1: Refactor bug. I used to have the date in the parameter list and removed it, not noticing that it was actually needed because of how WeekOfMonth is calculated.

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
BuddhaBuddy1: Yah, that is what let the previous bug get by. :-D

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
BuddhaBuddy1: Moving format character lookup to Initialize(), I didn't notice that section before. I am just saving the indexes in local private int fields. Let me know if there is a better way of doing that.

BuddhaBuddy1: I'm not sure if I understand your comments about batches. I don't really know how this all works in the background. Are you suggesting I actually cache indexFormat post-conversion for some period of time? I think moving the character lookups to Initialize() is the best I can do, and is worth the little bit of extra complexity.

I DO want to allow more than one special format character. I personally don't see the point in using more than one, but I don't want to limit others who might have different needs than I do. In fact, I previously limited to using only one of w, W, or WW, but technically, only W and WW are truly incompatible, so I now allow lowercase w and one of the upper case ones.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
karolz-ms: Thanks for changing this

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
karolz-ms: You can disregard my comments about batching etc. I like what you did here; it is simple and effective.

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
```diff
@@ -86,9 +88,40 @@ public void VerifyAllConfigOptionsMappedToConfigObject()
...
+        [InlineData(2023, 1, 11, "yyyy.MM.w", "2023.01.w2")]
+        [InlineData(2023, 2, 12, "yyyy.MM.w", "2023.02.w3")]
+        [InlineData(2023, 2, 1, "yyyy.W", "2023.w5")]
+        [InlineData(2023, 2, 1, "yyyy.WW", "2023.w05")]
+        [InlineData(2023, 4, 1, "yyyy.W", "2023.w13")]
+        [InlineData(2023, 4, 1, "yyyy.WW", "2023.w13")]
+        [InlineData(2023, 4, 2, "yyyy.WW", "2023.w14")]
```
karolz-ms, line 109: Consider adding a few more tests for more-than-one special replacement chars in a pattern. For example, with `yyyy.q.WW` as a pattern I would expect something like "2023.q1.w05" as the result of formatting 2/1/2023.

karolz-ms (approved):
Looks great, just one final suggestion. Thank you!

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
BuddhaBuddy1: Sorry, the proverbial fecal matter was hitting the rotating blades at work yesterday.
Sure, I'll get some more tests added.

BuddhaBuddy1: Finding the indexes early on causes issues because the position changes as I update the string. I either need to find positions live, or process the indexes in reverse order, highest found index first. I'll try that.

BuddhaBuddy1: I had to jump through some hoops to make sure this was still performant, but it is working now and checked in.

Honestly, if you would rather I rip it all out and just specify that only one special formatting character is allowed, I wouldn't push back, but it IS working. :-)

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -19,17 +21,43 @@
...
+            {
+                Index = index;
+                FormatIndexType = formatIndexType;
+            }
+
+            public int Index { get; set; }
+
+            public FormatIndexType FormatIndexType { get; set; }
+
+            public int CompareTo(object obj)
```
karolz-ms, line 51: One last minor thing: suggest use `IComparable<T>` instead of vanilla `IComparable` and return 1 if the passed object is null, see [this example](https://learn.microsoft.com/en-us/dotnet/api/system.icomparable-1?view=net-7.0#examples).

test/Microsoft.Diagnostics.EventFlow.Outputs.Tests/ElasticSearchOutputTests.cs
karolz-ms: No, no, I love it!

karolz-ms (approved):
Thank you @BuddhaBuddy1 ! Really appreciate you going through these iterations to make the change so nice.

karolz-ms:
@BuddhaBuddy1 you might have noticed that GH is giving us a warning sign for this PR, saying "branch is out of date with the base branch". This is because I have merged as small change to the build scripts while you were working on this PR.

I can fix it for you when merging the PR, so don't worry too much, but you can also fix it on your side by taking the GH offer to "update" your branch (merging the latest changes). There should be no conflicts--the latest changes are not touching the files you have been working on.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
karolz-ms: You know, scratch that. You put enough work into this. LMK help.

PR #418, opened 2023-02-17
Build/test fixes

karolz-ms:
Noticed our build scripts no longer work with latest MSBuild tooling. Fixed.
Also disabled running tests on .NET Core 3.1 since it is out of support.


xiaomi7732 (approved):
LGTM.

PR #419, opened 2023-02-22
Fix ElasticSearch index name formatting

karolz-ms:
A relatively minor fix to make all cases work as intended

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Diagnostics.EventFlow.Outputs
 
         private enum FormatIndexType {QuarterOfYear, WeekOfMonth, WeekOfYearDoubleDigit, WeekOfYear}
 
-        private class FormatIndex: IComparable<FormatIndex>
+        private class FormatIndex: IComparable<FormatIndex>, IEquatable<FormatIndex>
         {
             public FormatIndex(int index, FormatIndexType formatIndexType)
             {
@@ -54,6 +54,17 @@ namespace Microsoft.Diagnostics.EventFlow.Outputs
                 if (Index < other.Index) return 1;
                 return Index > other.Index? -1 : 0;
             }
+
+            public override bool Equals(object obj)
+            {
+                FormatIndex other = obj as FormatIndex;
+                if (other is null) return false;
+                return Index == other.Index;
+            }
+
+            public bool Equals(FormatIndex other) => Index == other.Index;
+
+            public override int GetHashCode() => Index.GetHashCode();
         }
 
         private FormatIndex[] formatIndexes;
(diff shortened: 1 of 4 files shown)


karolz-ms:
@BuddhaBuddy1 would appreciate a review as well--thx!

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
```diff
@@ -36,7 +36,7 @@ public class ElasticSearchOutput : IOutput
 
         private enum FormatIndexType {QuarterOfYear, WeekOfMonth, WeekOfYearDoubleDigit, WeekOfYear}
 
-        private class FormatIndex: IComparable
+        private class FormatIndex: IComparable<FormatIndex>
```
xiaomi7732, line 39: https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?redirectedfrom=MSDN&view=net-7.0#System_Object_Equals_System_Object_

Shall we follow this section:

Follow these guidelines when overriding [Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0#system-object-equals(system-object)):

Types that implement [IComparable](https://learn.microsoft.com/en-us/dotnet/api/system.icomparable?view=net-7.0) must override [Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0#system-object-equals(system-object)).

Types that override [Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0#system-object-equals(system-object)) must also override [GetHashCode](https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-7.0); otherwise, hash tables might not work correctly.

You should consider implementing the [IEquatable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-7.0) interface to support strongly typed tests for equality. Your [IEquatable<T>.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1.equals?view=net-7.0) implementation should return results that are consistent with [Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0).

If your programming language supports operator overloading and you overload the equality operator for a given type, you must also override the [Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0#system-object-equals(system-object)) method to return the same result as the equality operator. This helps ensure that class library code that uses [Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0) (such as [ArrayList](https://learn.microsoft.com/en-us/dotnet/api/system.collections.arraylist?view=net-7.0) and [Hashtable](https://learn.microsoft.com/en-us/dotnet/api/system.collections.hashtable?view=net-7.0)) behaves in a manner that is consistent with the way the equality operator is used by application code.

xiaomi7732 (approved):
Just one minor detail.

src/Microsoft.Diagnostics.EventFlow.Outputs.ElasticSearch/ElasticSearchOutput.cs
karolz-ms: That's a great point, thanks @xiaomi7732 ! Fixed now.

PR #420, opened 2023-05-09
Аdd all overrides for Write and WriteLine methods

DeBillieJoe:
Some packages use logging via log4net using various overrides and evenflow does not collect them

### DIFF
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
@@ -81,6 +81,36 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
             SubmitEventData(message, TraceEventType.Error);
         }
 
+        public override void Write(object o)
+        {
+            WriteLine(o.ToString());
+        }
+
+        public override void Write(object o, string category)
+        {
+            Write(o);
+        }
+
+        public override void Write(string message, string category)
+        {
+            WriteLine(message);
+        }
+
+        public override void WriteLine(object o)
+        {
+            WriteLine(o.ToString());
+        }
+
+        public override void WriteLine(object o, string category)
+        {
+            WriteLine(o);
+        }
+
+        public override void WriteLine(string message, string category)
+        {
+            WriteLine(message);
+        }
+
         public override void TraceData(TraceEventCache eventCache, string source, TraceEventType eventType, int id, object data)
         {
             if (this.Filter != null && !this.Filter.ShouldTrace(eventCache, source, eventType, id, null, null, data, null))
--- a/src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
+++ b/src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
@@ -88,12 +88,12 @@ namespace Microsoft.Diagnostics.EventFlow.Inputs
 
         public override void Write(object o, string category)
         {
-            Write(o);
+            WriteLine(o, category);
         }
 
         public override void Write(string message, string category)
         {
-            WriteLine(message);
+            WriteLine(message, category);
         }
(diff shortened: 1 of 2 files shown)


DeBillieJoe:
@microsoft-github-policy-service agree

karolz-ms:
@DeBillieJoe thank you for your contribution! Looks good; I just need to do some verification before merging and publishing, hope by the end of the week. Thanks again!

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
```diff
@@ -81,6 +81,36 @@ public override void Fail(string message, string detailMessage)
             SubmitEventData(message, TraceEventType.Error);
         }
 
+        public override void Write(object o)
+        {
+            WriteLine(o.ToString());
+        }
+
+        public override void Write(object o, string category)
```
karolz-ms, line 89: For this, and other overrides that take the `category` parameter, we should not lose the category information. This probably means adding another parameter to `SubmitEventData` so that the category information can be used to populate `EventData` object. For example:

```csharp
        private void SubmitEventData(string message, TraceEventType level, int? id = null, string source = null, string relatedActivityId = null, string category = null)
        {
            try
            {
                EventData eventEntry = new EventData()
                {
                    ProviderName = string.IsNullOrEmpty(source) ? TraceTag : source,
                    Timestamp = DateTimePrecise.UtcNow,
                    Level = ToEventLevel(level)
                };

                var eventPayload = eventEntry.Payload;
                eventPayload["Message"] = message;

                if (id != null && id.HasValue)
                {
                    eventPayload["EventId"] = id.Value;
                }

                if (!string.IsNullOrEmpty(relatedActivityId))
                {
                    eventPayload["RelatedActivityID"] = relatedActivityId;
                }

                if (!string.IsNullOrEmpty (category))
                {
                    eventPayload["Category"] = category;
                }

                // (rest of the SubmitEventData() code unchanged)
```

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
```diff
@@ -81,6 +81,36 @@ public override void Fail(string message, string detailMessage)
...
+        {
+            WriteLine(o);
+        }
+
+        public override void WriteLine(string message, string category)
+        {
+            WriteLine(message);
+        }
+
         public override void TraceData(TraceEventCache eventCache, string source, TraceEventType eventType, int id, object data)
```
karolz-ms, line 114: For these new overloads we should also add unit tests in https://github.com/Azure/diagnostics-eventflow/blob/master/test/Microsoft.Diagnostics.EventFlow.Inputs.Tests/TraceInputTests.cs. They can be very simple (similar to `TraceInputShouldOverrideWriteLine()`, just to verify that the override works and that the resulting `EventData` object passed down the pipeline is carrying the expected data.

karolz-ms (changes requested):
Thanks again for the contribution! I have a couple of suggestions

src/Microsoft.Diagnostics.EventFlow.Inputs.Trace/TraceInput.cs
karolz-ms: This implies for these overloads we would not rely on `WriteLine()`, but call `SubmitEventData()` directly.

DeBillieJoe:
I added the changes with [commit]([6496e63](https://github.com/Azure/diagnostics-eventflow/pull/420/commits/6496e635b303b2684e1ebf175dd6b3e7b938f4ae))

karolz-ms (approved):
Looks good

karolz-ms:
@DeBillieJoe the Trace input package with your changes is now on NuGet https://www.nuget.org/packages/Microsoft.Diagnostics.EventFlow.Inputs.Trace/1.6.1 🤗

Issue #402, opened 2022-04-22
Remove Nuget binary from sources


karolz-ms:
We have a Nuget binary checked in into our sources (`tools` folder). A better way would be to rely on a copy that exists in the build environment. 

The binary is used by `tools\restore.cmd` script. A straightforward attempt to replace the use with `dotnet restore` failed--the latter does not seem to properly restore the old-style signing projects that we use for binary and Nuget package signing.

xiaomi7732:
Lesson learned. @karolz-ms, have you tried to use this build step to bring in the binary:

```yaml
        - task: NuGetToolInstaller@1
          inputs:
            versionSpec: 
            checkLatest: true
```

PR #425, opened 2023-10-25
Refresh Azure.Identity dependency

karolz-ms:
Similar to https://github.com/Azure/diagnostics-eventflow/pull/424, but takes a newer version of the dependency, and also updates our NuGet package version, so that we can publish it on nuget.org.

Also updates NuGet.exe version.


xiaomi7732 (approved):
LGTM.
Could you please remind me, what is the reason for us to keep the binary of NuGet.exe?

karolz-ms:
@xiaomi7732 see https://github.com/Azure/diagnostics-eventflow/issues/402 for context

Issue #427, opened 2024-03-11
Superfluous dependencies to nuget packages


antymon4o:
The project Microsoft.Diagnostics.EventFlow.Core has superflous dependencies that are propagated to all other packages and projects that depend on it.
This includes all Microsoft.Diagnostics.EventFlow Inputs and Outputs packages.

List of packages:
1. System.Threading.Tasks.Parallel
The assembly System.Threading.Tasks.Parallel is included in .net standard 2.0 and later. For previous versions of .net standard a reference to the nuget package System.Threading.Tasks.Parallel is required. Therefore, only when tergeting .net standard 1.6 the reference to the nuget package is required. Otherwise, the dependency is superfluous.

2. Newtonsoft.Json
Core package does not have a functionality dependent on Newtonsoft.Json. There is a EventFlowJsonUtilities class that is use by some of 
the Outputs projects - EventHub, HttpOutput, Oms, StdOutput. Therefore, the dependency of Core to Newtonsoft.Json is superfluous. The negative impact is that, when some package refers Core it gets dependency to Newtonsoft.Json, even though it may never be needed.

PR #428, opened 2024-03-11
Superfluous dependencies to nuget packages

antymon4o:
This PR is related to #427.

Changes to EventFlow.Core Project are:
1. Sepcify dep to System.Threading.Tasks.Parallel only when targeting netstandard1.6
2. Remove dependency to Newtonsoft.Json
Extract classes that depend on Newtonsoft.Json to e new project
and add project dependency to it on the Outputs EventHub, HttpOutput, Oms, StdOutput.


src/Microsoft.Diagnostics.EventFlow/Microsoft.Diagnostics.EventFlow.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
karolz-ms, line 1: Could you please rename this project to `Microsoft.Diagnostics.EventFlow.Json`? It will reflect its purpose better.

src/Microsoft.Diagnostics.EventFlow.Outputs.EventHub/Microsoft.Diagnostics.EventFlow.Outputs.EventHub.csproj
```diff
@@ -20,6 +20,7 @@
 
   <ItemGroup>
```
karolz-ms, line 21: This, and the other outputs that will be using the new project, will also need a version change. A patch level upgrade is all that is needed probably.

src/Microsoft.Diagnostics.EventFlow/Microsoft.Diagnostics.EventFlow.csproj
```diff
@@ -0,0 +1,11 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard1.6;netstandard2.0;net462;net471</TargetFrameworks>
```
karolz-ms, line 4: Do we need all of these target frameworks? Or can we get away with, for example, only specifying `netstandard1.6` here?

karolz-ms (changes requested):
@antymon4o thank you for taking care of this issue, super appreciated!

Just a few minor suggestions/questions

src/Microsoft.Diagnostics.EventFlow/Microsoft.Diagnostics.EventFlow.csproj
antymon4o: In the class `MemberInfoConverter` there are conditional compilation directives. 

For two cases of `NETSTANDARD1_6` and `NET462 || NET471 || NETSTANDARD2_0` there are differences in the definition of the class and the methods that are overridden. For example, in the former case `MemberInfoConverter` derives from `JsonConverter`, but in the latter - from `JsonConverter<MemberInfo>`.

I am not sure what was the purpose of this approach, so I think it is more safe to leave `TargetFrameworks` and `MemberInfoConverter` as they are.

karolz-ms: Sounds good. I was just wondering.

karolz-ms (approved):

karolz-ms:
Updated packages have been published on Nuget. Thanks again.

